PR #4, opened 2021-03-18
SteamApi stuff isolated (may require some additional work)

espkk:
Needs review; DoD:
1) (BMS team) steamapi regression test
2) (anyone) general regression check


src/libs/SteamApi/src/achievements.cpp
```diff
@@ -1,5 +1,5 @@
 //============================================================================================
-#include "achievements.h"
+#include "achievements.hpp"
```
np-c0mp1ete, line 2: we don't use hpp extension in the project. do we really want to introduce it?

src/libs/SteamApi/.clang-tidy.cpt_backup
```diff
@@ -0,0 +1 @@
...ctors,modernize-use-uncaught-exceptions,modernize-use-using,readability-avoid-const-params-in-decls,readability-braces-around-statements,readability-const-return-type,readability-container-size-empty,readability-convert-member-functions-to-static,readability-deleted-default,readability-delete-null-pointer,readability-else-after-return,readability-identifier-naming,readability-implicit-bool-conversion,readability-inconsistent-declaration-parameter-name,readability-isolate-declaration,readability-magic-numbers,readability-misleading-indentation,readability-misplaced-array-index,readability-named-parameter,readability-non-const-parameter,readability-redundant-control-flow,readability-redundant-declaration,readability-redundant-function-ptr-dereference,readability-redundant-member-init,readability-redundant-preprocessor,readability-redundant-smartptr-get,readability-redundant-string-cstr,readability-redundant-string-init,readability-simplify-boolean-expr,readability-simplify-subscript-expr,readability-static-accessed-through-instance,readability-static-definition-in-anonymous-namespace,readability-string-compare,readability-uniqueptr-delete-release,readability-uppercase-literal-suffix'
```
np-c0mp1ete: why are these 3 .clang-tidy files required?

src/libs/SteamApi/src/achievements.cpp
espkk: Yes, I think we do. And all headers should be normalized to hpp.
however, here I did that unintentionally. But I still do not see much problem considering the above.
Please, let me know if we want to avoid this for now.

src/libs/SteamApi/.clang-tidy.cpt_backup
espkk: Good point. I believe this is clang tools' lack of manners. 
To be removed

ugeen4 (approved):

PR #5, opened 2021-03-18
Ini line endings fix

np-c0mp1ete:
Fix the issue with localized strings when you use english - you lose 1 char, when you use russian - you lose 1/2 of char and your string is not valid anymore


q4a:
This PR fixed my problem. Thanks.

espkk (approved):

PR #6, opened 2021-03-20
Remove threaded function queue

Hammie:
Started replacing the threaded implementation with standard threading code, but it turns out threaded queue is not currently used.

If we intend to use the queue, I'll add proper locking and signaling. Otherwise we could just get rid of this code altogether.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
-﻿build
+﻿/build/
+/cmake-build-*/
+
 .vs
+.idea
+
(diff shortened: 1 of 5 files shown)


espkk:
Seems fine so far, do you plan to do more commits?

Hammie:
> 
> 
> Seems fine so far, do you plan to do more commits?

It's finished. Did some more testing and this fixed my crashes. Debugger is happy again :)

espkk (approved):
+2

PR #7, opened 2021-03-20
Put logs & dump files to My Games\Sea Dogs folder. Pt 1

kk-se:
This commit consists of:
1. Put system.log, compile.log and error.log in user's Documents folders (logs in Sea Dogs\logs subfolder and dump file in Sea Dogs folder).
2. No common_h\rand.cpp, in common_h\rand.h are inline functions, using C++ library <random>.
To be finished, if considered useful.


src/apps/ENGINE/Main.cpp
```diff
@@ -3,13 +3,16 @@
...
 #include "s_debug.h"
+#include "compiler.h"
 #include "steam_api.h"
 #include <crtdbg.h>
 #include <dbghelp.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <stdio.h>
 #include <tchar.h>
 
+#define DUMP_FILENAME "engine_dump.dmp"
```
np-c0mp1ete: I'd suggest avoiding using macros if possible. constexpr variable is better.

src/apps/ENGINE/compiler.h
```diff
@@ -176,6 +177,9 @@ class COMPILER : public VIRTUAL_COMPILER
     ATTRIBUTES *rAP;
 
   public:
+    std::shared_ptr<spdlog::sinks::basic_file_sink_mt> error_warning_sink;
+    std::shared_ptr<spdlog::logger> tracelog, errorlog, warninglog;
```
np-c0mp1ete: I'd suggest putting each variable declaration on a separate line.

src/apps/ENGINE/Main.cpp
kk-se: OK, I'll do that.

I'm also changing #ifndef … #define … #endif checks to #pragma once, although #pragma is, in theory, outside of standard. However, AFAIK, all modern compilers properly supports it.

src/apps/ENGINE/compiler.h
kk-se: Be done.

src/apps/ENGINE/Main.cpp
espkk: This is to be defined in the coding style guidelines for the project which are not exist yet
Personally, I prefer #pragma once

espkk:
Ideally, there should be no public variables at all. Loggers should be aqaired by kind of getter (either core method or global namespaced with **log** or smthng like that). It can even be namespaced class **error** within **log** namespace which will have overloaded **operator()** for convenience
However, it's probably out of scope for this PR so it seems fine for me!

src/apps/ENGINE/Main.cpp
kk-se: Strange, I made changes, I checked the code compiles, then opened GitHub Desktop – it reported about two different files, so I put it with the name “Update 1”, but I don't see that here…
Still have to learn much about that system. :(
PS. It seems to be here: https://github.com/storm-devs/storm-engine/commit/c99883c890dc968b36823ebd7b4df95807ff95ad
But why didn't it update it here?
I hope I won't mess/broke that repository,

espkk: This diff is outdated, you commit is present in the new diffs

src/apps/ENGINE/Main.cpp
```diff
@@ -3,13 +3,16 @@
...
 #include "s_debug.h"
+#include "compiler.h"
 #include "steam_api.h"
 #include <crtdbg.h>
 #include <dbghelp.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <stdio.h>
 #include <tchar.h>
 
+constexpr auto DUMP_FILENAME = "engine_dump.dmp";
```
espkk, line 14: @kk-se see, it's now here :)

src/apps/ENGINE/Main.cpp
kk-se: > This diff is outdated, you commit is present in the new diffs

Thank you, now I see it's present in Commits tab.

kk-se:
> Ideally, there should be no public variables at all. Loggers should be aqaired by kind of getter (either core method or global namespaced with **log** or smthng like that). It can even be namespaced class **error** within **log** namespace which will have overloaded **operator()** for convenience

Since I believe it's not complete yet (the other log file, like attributes dump, screenshots, and other things should be, in my opinion, moved to the Documents folder, so game files and folders remain read-only), I'll take your advice into account and make appropriate adjustments.

np-c0mp1ete (approved):

espkk (approved):

PR #11, opened 2021-03-25
file_service: use std::fstream and std::filesystem for basic IO

q4a:


src/apps/ENGINE/file_service.h
```diff
@@ -88,47 +88,30 @@ class FILE_SERVICE : public VFILE_SERVICE
...
-    BOOL _DeleteFile(const char *lpFileName) override;
-    BOOL _WriteFile(HANDLE hFile, const void *lpBuffer, uint32_t nNumberOfBytesToWrite,
-                    uint32_t *lpNumberOfBytesWritten) override;
-    BOOL _ReadFile(HANDLE hFile, void *lpBuffer, uint32_t nNumberOfBytesToRead, uint32_t *lpNumberOfBytesRead) override;
+    std::fstream _CreateFile(const char *filename, std::ios::openmode mode) override;
+    void _CloseFile(std::fstream &fileS) override;
+    void _SetFilePointer(std::fstream &fileS, std::streamoff off, std::ios::seekdir dir) override;
+    int _DeleteFile(const char *filename) override;
+    bool _WriteFile(std::fstream &fileS, const char *s, std::streamsize count) override;
+    bool _ReadFile(std::fstream &fileS, char *s, std::streamsize count) override;
```
np-c0mp1ete: I think `s` could stay a void* to avoid all the reinterpret_casts. Same for _WriteFile

src/apps/ENGINE/Core.cpp
```diff
@@ -508,49 +509,61 @@ void CORE::ProcessRealize()
...
-    fio->_CloseHandle(fh);
+    }
+    fio->_CloseFile(fileS);
     delete State_file_name;
 
     const auto len = strlen(file_name) + 1;
     State_file_name = static_cast<char *>(new char[len]);
     if (State_file_name == nullptr)
+    {
         throw std::exception();
```
espkk: new throws std::bad_alloc
there is no point testing it with nullptr

src/apps/ENGINE/Core.cpp
```diff
@@ -508,49 +509,61 @@ void CORE::ProcessRealize()
 bool CORE::SaveState(const char *file_name)
 {
     if (!file_name)
+    {
         return false;
```
espkk: i'd suggest throwing std::logic_error here

src/apps/ENGINE/file_service.cpp
```diff
@@ -40,53 +41,54 @@ FILE_SERVICE::~FILE_SERVICE()
...
+    fileS.seekp(off, dir);
 }
 
-BOOL FILE_SERVICE::_DeleteFile(const char *lpFileName)
+int FILE_SERVICE::_DeleteFile(const char *filename)
 {
-    std::wstring filePathW = utf8::ConvertUtf8ToWide(lpFileName);
-    return DeleteFile(filePathW.c_str());
+    std::filesystem::path path = std::filesystem::u8path(filename);
+    return std::remove(filename);
```
espkk: you removing 'filename' instead of 'path'
I suppose you meant `std::filesystem::remove(path);`

src/apps/ENGINE/file_service.cpp
```diff
@@ -40,53 +41,54 @@ FILE_SERVICE::~FILE_SERVICE()
...
-    //    if(dwR != nNumberOfBytesToWrite) if(Exceptions_Mask & _X_NO_FILE_WRITE) throw
-    //    std::exception(_X_NO_FILE_WRITE);
-    return bRes;
+    fileS.exceptions(std::fstream::failbit | std::fstream::badbit);
+    try
+    {
+        fileS.write(s, count);
+        return true;
+    }
+    catch (std::fstream::failure e)
```
espkk: use
`catch (const std::fstream::failure& e)`
and add some logging about failed write

src/apps/ENGINE/file_service.cpp
```diff
@@ -40,53 +41,54 @@ FILE_SERVICE::~FILE_SERVICE()
...
-        *lpNumberOfBytesRead = dwR;
-    //    if(dwR != nNumberOfBytesToRead) if(Exceptions_Mask & _X_NO_FILE_READ) throw std::exception(_X_NO_FILE_READ);
-    return bRes;
+    fileS.exceptions(std::fstream::failbit | std::fstream::badbit);
+    try
+    {
+        fileS.read(s, count);
+        return true;
+    }
+    catch (std::fstream::failure e)
```
espkk: use
catch (const std::fstream::failure& e)
and add some logging about failed read

src/apps/ENGINE/file_service.cpp
```diff
@@ -170,17 +172,19 @@ BOOL FILE_SERVICE::_SetFileAttributes(const char *lpFileName, uint32_t dwFileAtt
...
+        _CloseFile(fileS);
         return nullptr;
     }
-    _CloseHandle(fh);
-    fh = _CreateFile(file_name, GENERIC_WRITE, 0, CREATE_NEW);
-    if (fh == INVALID_HANDLE_VALUE)
+    _CloseFile(fileS);
+    fileS = _CreateFile(file_name, std::ios::binary | std::ios::out);
+    if (!fileS.is_open())
+    {
```
espkk, line 186: please insert logging

src/apps/ENGINE/file_service.cpp
```diff
@@ -277,21 +281,25 @@ BOOL FILE_SERVICE::LoadFile(const char *file_name, char **ppBuffer, uint32_t *dw
     if (ppBuffer == nullptr)
         return false;
 
-    auto *const hFile = _CreateFile(file_name);
-    if (INVALID_HANDLE_VALUE == hFile)
+    auto fileS = fio->_CreateFile(file_name, std::ios::binary | std::ios::in);
+    if (!fileS.is_open())
```
espkk, line 288: we probably need logging here with level "verbose"

src/apps/ENGINE/ifs.cpp
```diff
@@ -362,38 +360,34 @@ bool IFS::VoidSym(char symbol)
...
-    if (fh == INVALID_HANDLE_VALUE)
-        return false;
-
-    const auto file_size = fs->_GetFileSize(fh, nullptr);
-    if (file_size == INVALID_FILE_SIZE)
+    }
+    auto fileS = fs->_CreateFile(_file_name, std::ios::binary | std::ios::in);
+    if (!fileS.is_open())
     {
-        fs->_CloseHandle(fh);
```
espkk, line 376: add some verbose logging?

espkk (changes requested):
good job! but please check these points

src/apps/ENGINE/file_service.cpp
q4a: Good catch. Thanks!

src/apps/ENGINE/Core.cpp
q4a: So I can just delete this check?

espkk: Exactly :)

src/apps/ENGINE/file_service.cpp
q4a: As far as I can see - there is no such level:
https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/common.h#L142

q4a: Ok, I got it: Trace=verbose

src/apps/ENGINE/ifs.cpp
q4a: There is file called `project.df` and as I understand it is some debug stuff and it's not necessary to have it. But I got error exception if I'll add
`core.tracelog->trace("Can not load file: %s", _file_name);`
here

np-c0mp1ete:
Thanks for your work! All good from my side.

espkk (approved):

PR #15, opened 2021-03-31
Fix geometry loading

Hammie:
Fixes a crash when trying to load a geometry file with a missing .col file

Also includes changes for #14

### DIFF
--- a/src/libs/Geometry/geom_static.cpp
+++ b/src/libs/Geometry/geom_static.cpp
@@ -8,7 +8,30 @@ Comments:
 Import library main file
 ******************************************************************************/
 #include "geom.h"
+
 #include <cstring>
+#include <vector>
+
+namespace {
+
+std::vector<unsigned int> getColData (GEOM_SERVICE &srv, const std::string_view& file_name) {
+    std::vector<unsigned int> result;
+
+    auto ltfl = srv.OpenFile(file_name.data());
+    if (ltfl.is_open()) {
+        const auto file_size = srv.FileSize(file_name.data());
+        if (file_size > 0)
+        {
+            result.resize(file_size / sizeof(int));
+            srv.ReadFile(ltfl, result.data(), file_size);
+        }
+    }
+    srv.CloseFile(ltfl);
+
+    return result;
+}
+
+} // namespace
 
 // create geometry func
 GEOS *CreateGeometry(const char *fname, const char *lightname, GEOM_SERVICE &srv, long flags)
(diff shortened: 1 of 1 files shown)


src/libs/Geometry/geom_static.cpp
```diff
@@ -150,21 +162,20 @@ GEOM::GEOM(const char *fname, const char *lightname, GEOM_SERVICE &_srv, long fl
     }
     srv.free(rvb);
     // read vertices
-    auto *_colData = colData;
-    if (lightname != nullptr && flsz / 4 != nvertices)
+    if (colData.size() != nvertices)
     {
-        srv.free(colData);
-        lightname = nullptr;
+        colData.clear();
```
espkk: Is there any sense reading colData before if we clear it here?
it seems like the following 'for' should just be wrapped into `if (colData.size() == nvertices)`

src/libs/Geometry/geom_static.cpp
```diff
@@ -150,21 +162,20 @@ GEOM::GEOM(const char *fname, const char *lightname, GEOM_SERVICE &_srv, long fl
...
     }
+    auto _colData = colData.begin();
```
espkk: Please, remove any underscore in the code you're modifying. this is a very bad style and should be avoided everywhere.
consider naming it colDataIt

src/libs/Geometry/geom_static.cpp
```diff
@@ -150,21 +162,20 @@ GEOM::GEOM(const char *fname, const char *lightname, GEOM_SERVICE &_srv, long fl
...
     {
         auto *vrt = static_cast<RDF_VERTEX0 *>(srv.LockVertexBuffer(vbuff[v].dev_buff));
         srv.ReadFile(file, vrt, vbuff[v].size);
         for (long vr = 0; vr < vbuff[v].nverts; vr++)
         {
             auto *prv = (RDF_VERTEX0 *)((uint8_t *)(vrt) + vbuff[v].stride * vr);
-            if (lightname != nullptr)
-                prv->color = *_colData++;
+            if (!colData.empty())
+                prv->color = *(_colData++);
```
espkk: This should be avoided if possible.
Shall be 
if (!colData.empty()) {
  prv->color = *_colData;
  ++_colData;
}

src/libs/Geometry/geom_static.cpp
```diff
@@ -19,20 +42,9 @@ GEOS *CreateGeometry(const char *fname, const char *lightname, GEOM_SERVICE &srv
...
-        flsz = srv.FileSize(lightname);
-        if (flsz > 0)
-        {
-            colData = static_cast<unsigned *>(srv.malloc(flsz));
-            srv.ReadFile(ltfl, colData, flsz);
-        }
-        else
-            lightname = nullptr;
-        srv.CloseFile(ltfl);
+    std::vector<unsigned int> colData;
```
espkk: please, use uint32_t (and add `#include <cstdint>`)
I see in the original code it was `unsigned int`, but let's fix that up when we go through

espkk (changes requested):
Thanks for the PR! Please, review the following points

espkk (approved):
Seems good to me, thanks

PR #17, opened 2021-04-01
Create py script to transform .gm to .obj

MangioneAndrea:


tools/gm-to-obj/README.md
```diff
@@ -0,0 +1,16 @@
+## .GM to .OBJ file converter
+
+### Requirements
+In order to run this script you need:
+- Python 3+
+- Some .gm files from one game
+
+### How to
+To run the script, just run
+``python .\gm-to-obj.py``
```
np-c0mp1ete: I think the command line interface should be `gm-to-obj.py model_path.gm [-o new_model_path.obj]`

if no output file is specified, then an obj file should be created with the same name (but different file extension) and in the same directory as an input file

tools/gm-to-obj/gm-to-obj.py
```diff
@@ -0,0 +1,298 @@
...
+    # Read an array of the struct type with the fiven size
+    res = []
+    for _ in range(0, size):
+        m = eval(struct)()
+        f.readinto(m)
+        res.append(m)
+    return res
+
+
+user_input = input("Enter path of the .gm file ")
```
np-c0mp1ete: AFAIK, it's considered a good practice to wrap the code in `if __name__ == '__main__':`

tools/gm-to-obj/README.md
```diff
@@ -0,0 +1,16 @@
+## .GM to .OBJ file converter
+
+### Requirements
+In order to run this script you need:
+- Python 3+
+- Some .gm files from one game
```
np-c0mp1ete: I have a problem converting RESOURCE/MODELS/light.gm. Please make sure that all gm files from Sea Dogs: To Each His Own can be successfully converted.

MangioneAndrea: I'm going to investigate why. GM Viewer can't open that file either...

tools/gm-to-obj/gm-to-obj.py
MangioneAndrea: By adding that, the script cannot be called by other scripts. Are you sure you want that?

np-c0mp1ete: I believe, argument parsing should be wrapped in `if __name__ == '__main__':` and the actual logic of a script should be in a function to allow calling from other scripts, as you said

tools/gm-to-obj/README.md
np-c0mp1ete: I would not be surprised, if there are a lot of gm files that are not used at all. Maybe some leftovers from the previous games. Maybe even the gm file format has changed since then. 
In the future we have to find it out to make the transition from gm to obj as smooth as possible. But I guess it's out of scope of this PR. 
So, just not crashing when processing a file and printing an error with some explanation what is wrong with a gm file would be good enough.

egorpugin:
Isn't it easier to write such thing in C/C++?

espkk:
I think we want to keep convertors and stuff in python.
They are:
1) portable
2) embeddable
3) easy to understand for those who don't know c++

tools/gm-to-obj/README.md
MangioneAndrea: Sure, I've also seen that a couple of ships crash the script as well, so there is surely a bug in my code :) 
I'm going to have a deeper look when I got some time

vorot93:
Why not glTF?
https://godotengine.org/article/we-should-all-use-gltf-20-export-3d-assets-game-engines

np-c0mp1ete:
> Why not glTF?
> https://godotengine.org/article/we-should-all-use-gltf-20-export-3d-assets-game-engines

That might be an interesting option for the future some day. At the moment, converting to obj seems like the easiest way to go.

tools/gm-to-obj/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+### How to
+To run the script, just type in the console
+``python .\gm-to-obj.py {filepath}`` 
+
+E.g. ``python .\gm-to-obj.py "E:\SteamLibrary\steamapps\common\Sea Dogs To Each His Own\RESOURCE\MODELS\fish01.gm" -o myCoolFish.obj``
+
+Option | Short | Description | Default
+------ | ----- | ----------- | -------
+``--output``|``-o``| The name of the ``.json`` file to create| Name of the input file with ``.json`` extension
```
np-c0mp1ete: json?

tools/gm-to-obj/gm-to-obj.py
```diff
@@ -0,0 +1,371 @@
...
+    # define args
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "path", help="Where is the .gm file located?", type=str)
+    parser.add_argument("--output", "-o", help="Output name", type=str)
+    args = parser.parse_args()
+    args_input = str(args.path)
+    args_output = args.output
+    if(args_output is None):
+        args_output = ntpath.basename(args_input).replace(".gm", ".obj")
```
np-c0mp1ete: I think output files should be written in the same directory as input files.
Running
`$ ./gm-to-obj.py ../../RESOURCE/MODELS/WorldMap/spanish_manowar.gm`
outputs a file in the current directory.

np-c0mp1ete (changes requested):
Thanks a lot, it got much more stable!
Still, if I run the script on all gm files, I get some python errors:
`find ../../RESOURCE/ -iname *.gm -exec python ./gm-to-obj.py \{\} \; > ./log.txt`
Could you please take a look at them?

Hammie:
I tried to convert Oxbay port from PotC using this script and ran into some issues.

I had to modify the script and change `c_long` and `c_ulong` to `c_int` and `c_uint`. I'm running a 64-bit version of python3 so that is probably related (not a python expert myself).

Aside from this issue I was able to convert the mesh, which I could not do with the old gm-viewer due to it having more then 32k vertices. The result however appears to be mirrored, as the pier should be on the left:

Oxbay port (mirrored)

The script has already been useful to me though, so thanks for making this!

MoffKalast:
Nice work, super useful. 👍

MangioneAndrea:
@Hammie Thanks for the feedback. I'm happy it could help :)

Is this problem just there or everywhere and which axis is it? If it's everywhere I will mirror everything (and add an option to avoid it :D)

About the long/int issue, that is weird, I do not own that game, so I cannot test it, on TEHO I've got no issue and I'm on a x64 system as well

tools/gm-to-obj/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+E.g. ``python .\convert_all.py "E:\SteamLibrary\steamapps\common\Sea Dogs To Each His Own\RESOURCE\MODELS\Ammo"``
+
+## Issues
+
+The following files will not convert, as they cannot be representend as .obj:
+
+TEHO
+
+- `big_sky.gm`
```
np-c0mp1ete, line 39: @mitrokosta any updates on whether these are used or not?
@MangioneAndrea if there are no other known issues with the script, I think we can merge it in its current state.

MangioneAndrea: There are those 2 problems metioned by @Hammie .The mirroring looks important, if it's not urgent to merge it, i'd like to dive into it first

MangioneAndrea: The x-axis was indeed inverted, at least for items in TEHO, now that problem is fixed.
About the other one, I cannot reproduce it, so I'd say that it can be merged now

np-c0mp1ete (dismissed):
Thank you for this great tool! There is still a lot of information that should be retrieved from gm files, but I think now we have a solid starting point for the next iterations of the gm-to-obj transition.

np-c0mp1ete:
@MangioneAndrea could you rebase the branch onto the upstream develop before I merge? Obviously, there are no conflicts, but it makes the git history cleaner.

espkk (dismissed):

np-c0mp1ete (approved):

PR #29, opened 2021-04-03
Add functions for comparing strings in a case-insensitive manner

Hammie:
I've also added a script I often use to simplify the configuration of libraries and executables.

Added Catch2 as a development dependency, as well as some unit tests to test the new string comparison functions. Unit tests have also been configured to be ran by the CI configuration.

Created a new "common" library we can use for things like utilities, my intent is to have it replace most of Common_h after the code has been cleaned up.

I'm not all to happy about the names of `iEquals` and `iLess`, so if anyone has better suggestions, please let me know.


src/libs/common/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+project(common)
```
espkk: I'd suggest renaming 'common' into 'util'
Their semantics are similar, though we already have common_h (which I want to get rid of though)
Also, 'util' reflects its contents slightly better

src/libs/common/test/string_compare.test.cpp
```diff
@@ -0,0 +1,75 @@
+#include "storm/string_compare.hpp"
+
+#include <catch2/catch.hpp>
+
+using namespace storm;
```
espkk: if this possible to move these two 'using namespace' inside TEST_CASE?
if so, do that, please

espkk (changes requested):
Thanks for the PR! Seems good to me overall, but please check the following points

src/libs/common/CMakeLists.txt
Hammie: It was called _utils_ at first, then I changed it to _common_, so no problem changing it back to _util_.

src/libs/common/test/string_compare.test.cpp
Hammie: sure :+1:

src/libs/common/CMakeLists.txt
espkk: 'utils' is maybe even better. it's up to you here :)

espkk (approved):

PR #30, opened 2021-04-04
inifile: return unique_ptr

vorot93:


espkk (approved):
Seems good to me. Thanks!

PR #32, opened 2021-04-04
file_service: fix save and load utf8 filenames

q4a:


src/apps/ENGINE/file_service.cpp
```diff
@@ -41,7 +41,8 @@ FILE_SERVICE::~FILE_SERVICE()
 
 std::fstream FILE_SERVICE::_CreateFile(const char *filename, std::ios::openmode mode)
 {
-    std::fstream fileS(filename, mode);
+    std::filesystem::path path = std::filesystem::u8path(filename);
```
np-c0mp1ete, line 44: u8path is deprecated in C++20. Should we stop using it?

q4a: I think that it's ok for now. But in future we will have to replace all `char` to `char8_t` and `string` to `u8string`:
https://en.cppreference.com/w/cpp/language/types#char8_t 
https://en.cppreference.com/w/cpp/string/basic_string

espkk (approved):
Thanks for the PR!

PR #40, opened 2021-04-09
Replacement for last Windows-only file_system methods

q4a:
Replace _CreateDirectory, _GetCurrentDirectory, _SetCurrentDirectory using std::filesystem analogs
Replace _GetExecutableDirectory with SDL_GetBasePath

### DIFF
--- a/src/apps/ENGINE/file_service.cpp
+++ b/src/apps/ENGINE/file_service.cpp
@@ -187,8 +187,7 @@ void FILE_SERVICE::_FlushFileBuffers(std::fstream &fileS)
 std::string FILE_SERVICE::_GetCurrentDirectory()
 {
     const auto curPath = std::filesystem::current_path().u8string();
-    std::string result(curPath.size(), '\0');
-    std::memcpy(result.data(), curPath.data(), curPath.size());
+    std::string result(curPath.begin(), curPath.end());
     return result;
 }
 
--- a/src/apps/ENGINE/file_service.cpp
+++ b/src/apps/ENGINE/file_service.cpp
@@ -6,6 +6,7 @@
 #include <exception>
 #include <storm/string_compare.hpp>
 #include <string>
+#include <SDL.h>
 
 #define COMMENT ';'
 #define SECTION_A '['
@@ -56,7 +57,7 @@ void FILE_SERVICE::_SetFilePointer(std::fstream &fileS, std::streamoff off, std:
     fileS.seekp(off, dir);
 }
 
-int FILE_SERVICE::_DeleteFile(const char *filename)
+bool FILE_SERVICE::_DeleteFile(const char *filename)
 {
     std::filesystem::path path = std::filesystem::u8path(filename);
     return std::filesystem::remove(path);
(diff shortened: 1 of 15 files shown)


src/apps/ENGINE/ifs.cpp
```diff
@@ -432,9 +432,9 @@ void IFS::Format(char *file_data, long file_size)
         }
     }
 
-    for (n = 0; n < file_size; n++)
+    for (n = -1; n < file_size; n++)
```
np-c0mp1ete: This fix is in a separate PR.

np-c0mp1ete (changes requested):
Thanks for the PR! 

Please remove the ifs.cpp fix from this PR and try to add sdl2 via conan or as a submodule in the worst case.

src/apps/ENGINE/ifs.cpp
q4a: > Thanks for the PR!
> 
> Please remove the ifs.cpp fix from this PR and try to add sdl2 via conan or as a submodule in the worst case.

I found way to get sdl2 via conan, but I have another question: you would like to link sdl2 statically or dynamically?

q4a:
> Thanks for the PR!
> 
> Please remove the ifs.cpp fix from this PR and try to add sdl2 via conan or as a submodule in the worst case.

I found way to get sdl2 via conan, but I have another question: you would like to link sdl2 statically or dynamically?

src/apps/ENGINE/file_service.h
```diff
@@ -101,14 +101,13 @@ class FILE_SERVICE : public VFILE_SERVICE
     std::time_t _ToTimeT(std::filesystem::file_time_type tp) override;
     std::filesystem::file_time_type _GetLastWriteTime(const char *filename) override;
     void _FlushFileBuffers(std::fstream &fileS) override;
-    uint32_t _GetCurrentDirectory(uint32_t nBufferLength, char *lpBuffer) override;
```
np-c0mp1ete, line 104: I think this function should just return std::string. Passing a buffer to fill without passing the length of this buffer is a bad idea imo

np-c0mp1ete (changes requested):

src/apps/ENGINE/file_service.h
q4a: Done.

src/apps/ENGINE/file_service.cpp
```diff
@@ -172,27 +173,18 @@ void FILE_SERVICE::_FlushFileBuffers(std::fstream &fileS)
...
+std::string FILE_SERVICE::_GetCurrentDirectory()
 {
-    wchar_t BufferW[MAX_PATH];
-    uint32_t Res = GetCurrentDirectory(nBufferLength, BufferW);
-    std::string CurrentDirectory = utf8::ConvertWideToUtf8(BufferW);
-    strcpy_s(lpBuffer, nBufferLength, CurrentDirectory.c_str());
-    return Res;
+    const auto curPath = std::filesystem::current_path().u8string();
+    std::string result(curPath.size(), '\0');
+    std::memcpy(result.data(), curPath.data(), curPath.size());
```
np-c0mp1ete: This can be done similar to how @Hammie proposed for the other PR.
`std::string result(curPath.begin(), curPath.end());`

src/apps/ENGINE/s_dbg_sourceview.cpp
```diff
@@ -533,7 +533,7 @@ bool SOURCE_VIEW::OpenSourceFile(const char *_filename)
     }
 
     char DirectoryName[MAX_PATH];
-    fio->_GetCurrentDirectory(sizeof(DirectoryName), DirectoryName);
+    strcpy(DirectoryName, fio->_GetCurrentDirectory().c_str());
```
np-c0mp1ete: this seems fragile. MAX_PATH is only 256 and with Russian symbols in the directory name that can easily go out of bounds. I'd suggest replacing char arrays with std::string here and in other similar places

np-c0mp1ete (changes requested):

src/apps/ENGINE/s_dbg_sourceview.cpp
espkk: MAX_PATH is NOT for UTF-8. As far as I remember, UTF8 path may be up to 32k bytes long

src/apps/ENGINE/file_service.cpp
q4a: fixed

src/apps/ENGINE/s_dbg_sourceview.cpp
q4a: I used MAX_PATH *4 for UTF8 path: https://stackoverflow.com/questions/7106911/problem-with-handling-path-length

src/apps/ENGINE/s_dbg_sourceview.cpp
```diff
@@ -532,8 +532,8 @@ bool SOURCE_VIEW::OpenSourceFile(const char *_filename)
         CDebug.SaveRecentFileALine(SourceFileName, nActiveLine);
     }
 
-    char DirectoryName[MAX_PATH];
-    fio->_GetCurrentDirectory(sizeof(DirectoryName), DirectoryName);
+    char DirectoryName[MAX_PATH * 4];
```
np-c0mp1ete: it doesn't seem like a proper solution. I'd propose to replace char[] with std::string

q4a: Check new version please.

np-c0mp1ete (approved):

PR #41, opened 2021-04-09
Use SDL for window and event handling.

ObKo:
This PR Introduces window and OS event handling abstraction - WindowManager class. And its implementation using SDL.

WindowManager defines OS-agnostic event struct (WMEvent), keyboard keys, etc. It handle OS events and can forward it to clients (Core, PCS_Control) using callbacks. Also, it provides methods to request current input state, replacing WinAPI's GetAsyncKeyState()

Currently, it uses WinAPI's virtual key (VK_*) for keycodes for compatibility and converts them from/to SDL scancodes (aka "USB scancodes"). This conversion can be eliminated if we replace keycodes everywhere in code, **including scripts** which is work for future.

pcs_controller and Core updated to use this new class. Core initialization code is messed, so WM is initialized outside of it for now.

Tested for some time on TEHO - seems everything (sailing, walk, battle, gui) works same as before,

Feel free to criticize this new approach.

P.S.: There is https://github.com/storm-devs/storm-engine/pull/40 which also adds SDL, I wil rebase this PR when we will decide what to do with SDL.


src/apps/ENGINE/Main.cpp
```diff
@@ -11,6 +11,8 @@
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/spdlog.h>
 
+#include "../../libs/SDLWindowManager/SDLWindowManager.hpp"
```
np-c0mp1ete: since SDLWindowManager lib is in the include path of the engine target, you could just write
`#include <SDLWindowManager.hpp>`

src/libs/Common_h/CMakeLists.txt
```diff
@@ -89,6 +89,7 @@ set(PROJECT_SOURCES
     vmodule_api.h
     vparticle_system.h
     Weather_Base.h
+	WindowManager.hpp
```
np-c0mp1ete: We want to get rid of the common_h library at some point, it's a rudiment of the plugin system (each library was linked dynamically as a plugin).
So, I'd propose moving WindowManager.hpp to the SDLWindowManager library. Probably makes sense to remove SDL from the name too

src/libs/renderer/sdevice.cpp
```diff
@@ -453,7 +453,7 @@ bool DX9RENDER::Init()
         for (long i = 0; i < iCapBuffers; i++)
             aCaptureBuffers.push_back(new char[sizeof(uint32_t) * screen_size.x * screen_size.y]);
 
-        if (bWindow)
+        /*if (bWindow)
```
np-c0mp1ete: Should it be commented out?

GovorovViva64:
Hello @ObKo, did you have to insert the whole SDL2 sources?
Could this be done by using a Conan dependency management or Git submodules?

np-c0mp1ete:
@GovorovViva64, yes, we want to do exactly this in https://github.com/storm-devs/storm-engine/pull/40. This branch will be rebased and merged after.

src/libs/SDLWindowManager/SDLWindowManager.hpp
```diff
@@ -0,0 +1,32 @@
...
+    void eventUnsubscribe(int id) override;
+
+    void processEvents() override;
+
+    bool keyboardKeyState(const KeyboardKey &key) override;
+
+    bool mouseKeyState(const MouseKey &key) override;
+
+  private:
+    std::list<std::pair<int, WMEventHandler>> m_handlers;
```
np-c0mp1ete: why not std::vector?

src/libs/Common_h/WindowManager.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+//! Mouse keys
+enum class MouseKey
+{
+    Left,
+    Right,
+    Middle
+};
+
+//! Window/input event
```
np-c0mp1ete: Why `!` in front of the comments?

src/libs/Common_h/WindowManager.hpp
```diff
@@ -0,0 +1,89 @@
+#pragma once
+
+#include <functional>
+#include <variant>
+
+namespace storm
+{
+// TODO: Make own type for keycodes and use it everywhere (reuse SDL scancodes?)
+//! Keyboard key, using virtual key codes (VK_* frim WinUser.h) for now
+typedef unsigned int KeyboardKey;
```
np-c0mp1ete: I'd suggest using `using` instead of `typedef`

src/libs/Common_h/controls.h
```diff
@@ -184,16 +194,11 @@ class CONTROLS
     virtual void SetMouseSensivityY(float){};
     virtual short GetDebugAsyncKeyState(int vk)
     {
-        return GetAsyncKeyState(vk);
+        return 0;
```
np-c0mp1ete: Should the functionality be restored here and in GetDebugKeyState?

src/libs/SDLWindowManager/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+project(SDLWindowManager)
+
+find_package(SDL2 REQUIRED)
+
+add_library(${PROJECT_NAME} STATIC
+    SDLWindowManager.hpp
```
np-c0mp1ete: in the new code we put public headers in the `include` subfolder. See https://github.com/storm-devs/storm-engine/tree/develop/src/libs/SteamApi

src/libs/SDLWindowManager/SDLWindowManager.cpp
```diff
@@ -0,0 +1,420 @@
...
+#include <SDL_syswm.h>
+#include <SDL_video.h>
+#include <map>
+
+namespace storm
+{
+namespace
+{
+// TODO: Use scancodes everywhere instead of Windows virtual key
+static const std::map<unsigned int, unsigned int> VK_TO_SDL_MAP{
```
np-c0mp1ete: It seems, window management and input management got mixed into one class. I guess it's better to keep them separated

np-c0mp1ete (commented):
Thanks for the PR! My two cents:
I think, we want to split window management from input handling even though they are somewhat connected.
I'd try to make an abstraction layer on top of input and window managers to hide from the rest of the codebase the fact that they use sdl internally. So, no names containing "sdl" in their public APIs.
These are just my ideas, feel free to disagree.

src/libs/renderer/sdevice.cpp
ObKo: This code just force-moves window to center of screen using non-portable WinAPI. I see no reason to keep this code, because SDL centers window by default.

src/libs/Common_h/CMakeLists.txt
ObKo: WindowManager is abstract class with own (shared) header, SDLWindowManager - one of possible implementation of WindowManager. I thought that Common_h is some kind of "include" directory - where all interfaces (abstract classes) are stored.

src/libs/SDLWindowManager/SDLWindowManager.hpp
ObKo: Because handlers can be dynamically added/removed in arbitrary order...
But this insertion/removes are rare so maybe std::vector will suffice here.

src/libs/Common_h/WindowManager.hpp
ObKo: It's for documentation tools, like doxygen https://www.doxygen.nl/manual/docblocks.html 
Visual Studio also supports this

src/libs/Common_h/controls.h
ObKo: CONTROLS is also abstract class, which isn't used directly. Removed this to get rid of WinAPI calls.

Hammie: Maybe make this a pure virtual function?

espkk:
Thanks for the PR! Is it ready for review now?

ObKo:
Yes

src/libs/Input/include/Input.hpp
```diff
@@ -0,0 +1,124 @@
+#pragma once
+
+#include <functional>
+#include <variant>
+
+namespace storm
+{
+// TODO: Make own type for keycodes and use it everywhere (reuse SDL scancodes?)
+//! Keyboard key, using virtual key codes (VK_* frim WinUser.h) for now
+using KeyboardKey = unsigned int;
```
np-c0mp1ete: maybe use fixed width types for input values, e.g. uint32_t?

src/libs/Input/include/Input.hpp
```diff
@@ -0,0 +1,124 @@
+#pragma once
+
+#include <functional>
+#include <variant>
+
+namespace storm
+{
+// TODO: Make own type for keycodes and use it everywhere (reuse SDL scancodes?)
+//! Keyboard key, using virtual key codes (VK_* frim WinUser.h) for now
```
np-c0mp1ete:
```suggestion
//! Keyboard key, using virtual key codes (VK_* from WinUser.h) for now
```

src/libs/Input/include/Input.hpp
```diff
@@ -0,0 +1,124 @@
...
+    //! \return true if key is pressed
+    virtual bool KeyboardKeyState(const KeyboardKey &key) const = 0;
+
+    //! Current mouse key state
+    //! \return true if key is pressed
+    virtual bool MouseKeyState(const MouseKey &key) const = 0;
+
+    //! Current controller button state
+    //! \return true if button is pressed
+    virtual bool ControllerButtonState(const ControllerButton &button) const = 0;
```
np-c0mp1ete, line 118: `ControllerButtonState()` and `ControllerAxisValue()` are unused. That's fine, I don't mind leaving it as it is. Just wondering, do you have plans to continue working on the controller support in the future?

src/libs/Input/src/SDLInput.cpp
```diff
@@ -0,0 +1,456 @@
...
+    keyStates_ = SDL_GetKeyboardState(nullptr);
+    SDL_SetRelativeMouseMode(SDL_TRUE);
+    OpenController();
+}
+
+int SDLInput::Subscribe(const EventHandler &handler)
+{
+    int id = 1;
+    if (!handlers_.empty())
+        id = (--handlers_.end())->first + 1;
```
Hammie:
```suggestion
        id = handlers_.rbegin()->first + 1;
```

src/libs/Input/include/Input.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+namespace storm
+{
+// TODO: Make own type for keycodes and use it everywhere (reuse SDL scancodes?)
+//! Keyboard key, using virtual key codes (VK_* frim WinUser.h) for now
+using KeyboardKey = unsigned int;
+
+//! Mouse relative position
+// FIXME: Better type for that?
+using MousePos = std::pair<int, int>;
```
Hammie: Using a simple struct should do the trick?
```suggestion
struct MousePos {
    int x{};
    int y{};
};
```

src/libs/Input/include/Input.hpp
```diff
@@ -0,0 +1,124 @@
...
+        ControllerButtonUp    //!< Controller button released, data contains ControllerButton
+    };
+    //! Event type
+    Type type;
+    //! Data associated with event
+    std::variant<std::string, KeyboardKey, MousePos, MouseKey, ControllerAxisState, ControllerButton> data;
+};
+
+//! Abstract input manager that handles all input events
+class Input
```
Hammie, line 92: Should have a virtual destructor

src/libs/Window/include/OSWindow.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <functional>
+
+namespace storm
+{
+//! Abstract window
+class OSWindow
```
Hammie, line 8: Should have a virtual destructor

src/apps/ENGINE/Main.cpp
```diff
@@ -97,143 +107,80 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
...
-    wndclass.lpszClassName = windowName;
-    wndclass.hIconSm = LoadIcon(nullptr, IDI_APPLICATION);
-    RegisterClassEx(&wndclass);
-
-    auto *const hwnd = CreateWindow(windowName, windowName, WS_POPUP, 0, 0, 0, 0, NULL, NULL, hInstance, NULL);
-    ShowWindow(hwnd, SW_SHOWNORMAL);
+    std::shared_ptr<storm::OSWindow> window = storm::OSWindow::Create(width, height, fullscreen);
+    window->SetTitle("Sea Dogs");
+    core.Set_Hwnd(static_cast<HWND>(window->OSHandle()));
+    auto windowHandler = [&](const storm::OSWindow::Event &e) {
```
Hammie: I would outline this lambda into a proper function to try and keep the size of WinMain under control

Hammie (changes requested):
I don't really like how SDLWindow::windows keeps a static list of all windows, I'll think about it some more but I think there should be a better solution.

I also think we may be taking the abstraction a bit far with Input and OSWindow, in particular I would get rid of the `Create` functions. 

Base classes should also always have virtual destructors.

np-c0mp1ete (commented):
Thanks for the PR! I like how the input code is now separated from the window handling. Please don't forget to rebase the branch after resolving @Hammie's comments.

src/libs/Input/include/Input.hpp
ObKo: Yes, but I should have deep dive into controls subsystem first

ObKo:
> I also think we may be taking the abstraction a bit far with Input and OSWindow, in particular I would get rid of the `Create` functions.

How should then I create SDLWindow without using "private" SDLWindow header? Or this can be considered acceptable?

Hammie (approved):
Looks good to me. Thanks!

np-c0mp1ete (approved):
Thanks for the final improvements! Please rebase your branch and I'll merge it.

PR #47, opened 2021-04-10
Finalize (for now) Scripting Overview article

Yakvi:
Task card: https://github.com/storm-devs/storm-engine/projects/4#card-58081815

Expand Scripting Overview article as follows:
* Add information on scripting types. 
* Add information on branching.
* Add expanded information on arrays initialization. (might move out later to a "special call outs" article.)
* Add misc information.

### DIFF
--- a/kb/0002-scripting-overview.md
+++ b/kb/0002-scripting-overview.md
@@ -40,6 +40,147 @@ The following list is sorted alphabetically and is based on the latest build of 
 * `program\weather\` - weather system definitions and usage.
 * `program\worldMap\` - everything world map related.
 
+## Typing
+
+In order to be compiled, the scripting files should contain good C-like code. This also means that the variables should be strongly typed. Commonly used are base C types: `int`, `float`, but also `bool` and a simple to use `string`. Any custom structs are supported in the form of `object`s. Last, arrays are supported of the above.
+
+### Base Types
+
+There're three base types in scripting language: 
+
+* `int`: a signed 32-bit integer
+
+``` C++
+syntax: 
+	int var_name;
+examples:
+	int GameTime = -2053; // OK
+	int FreeSlots = 75;   // OK
+```
+
+* `float`: a single-precision floating-point number. Because there're no `double`s, specifying `f` at the end is not required but allowed. 
+
+``` C++
+syntax: 
+	float var_name;
+examples:
+	float WindSpeed = 2.0;         // OK
+	float WindDirection = -0.0057; // OK
+	float ShipDirection = 3.14f;   // OK
+```
+
+* `bool`: a logical `true`/`false` operator where `true` is any non-zero value and `false` is zero.
+
+``` C++
+syntax: 
+	bool var_name;
+examples:
+	bool IsSpeaking = true;                  // OK
+    bool IsDrunk = (AlcoholInBlood > 10.0f); // OK
+    bool KilledAtLeastOneEnemy = KillCount;  // OK where KillCount is an int
+```
+
+### Strings
+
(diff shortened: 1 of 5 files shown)


kb/0002-scripting-overview.md
```diff
@@ -40,6 +40,461 @@ The following list is sorted alphabetically and is based on the latest build of
 * `program\weather\` - weather system definitions and usage.
 * `program\worldMap\` - everything world map related.
 
+## Typing
+
+In order to be compiled, the scripting files should contain good C-like code. This also means that the variables should be strongly typed. Commonly used are base C types: `int`, `float`, but also `bool` and a simple to use `string`. Any custom structs are supported in the form of `object`s. Last, arrays are supported of the above.
```
np-c0mp1ete: > Last, arrays are supported of the above.

Is it a correct English sentence?

kb/0002-scripting-overview.md
```diff
@@ -40,6 +40,461 @@ The following list is sorted alphabetically and is based on the latest build of
...
+```
+
+* `bool`: a logical `true`/`false` operator where `true` is any non-zero value and `false` is zero.
+
+``` C++
+syntax: 
+	bool var_name;
+examples:
+	bool IsSpeaking = true;                  // OK
+    bool IsDrunk = (AlcoholInBlood > 10.0f); // OK
```
np-c0mp1ete, line 81: Indentation in not consistent here and in the `strings` section

kb/0002-scripting-overview.md
```diff
@@ -40,6 +40,461 @@ The following list is sorted alphabetically and is based on the latest build of
...
+example:
+	int Characteristics[10];                // OK
+	string CharacteristicsNames[10];        // OK
+
+	int array_size;
+	array_size = 10;
+	string CharacteristicsNames[array_size]; // Also OK
+```
+
+// TODO(yakvi): Info below is dated 2009. Is this still relevant?
```
np-c0mp1ete: @mitrokosta Could you help us out here, is it still the case?

kb/0002-scripting-overview.md
```diff
@@ -40,6 +40,461 @@ The following list is sorted alphabetically and is based on the latest build of
...
+        int ExampleFour[3];
+        ExampleThree[0] = 0;                               // OK
+        ExampleThree[1] = 5;                               // OK
+        ExampleThree[2] = 10;                              // OK
+    }
+```
+
+### Objects, References and Attributes
+
+An `object` is a treelike text structure of an arbitrary shape. This means that it can store any amount of fields in _text_ form, even if specifying a number without quotes is allowed.
```
np-c0mp1ete: > even if specifying a number without quotes is allowed

`even though` fits better imo.

np-c0mp1ete (commented):
Thanks a lot for taking care of the documentation!

kb/0002-scripting-overview.md
mitrokosta: Most likely yes. According to [this](https://github.com/storm-devs/storm-engine/commit/76e8eaf9cde9ed95dc5b65e34e57a3613d387242) commit's diff compiler's array handling mechanism was not changed.

kb/0002-scripting-overview.md
Yakvi: Could be correct but a bit wanky indeed. Will rework.

Yakvi:
Feedback implemented. Thanks for the quick turnaround!

kb/0002-scripting-overview.md
Yakvi: I didn't realize that in the original documentation tabs were used instead of spaces. Fixed!

PR #49, opened 2021-04-11
Programming style guide

ObKo:
Add initial style guide, based on https://github.com/lefticus/cppbestpractices

It is quite outline, so we should add more сlarifications (strings, smart pointers, etc.) later.

### DIFF
--- a/kb/0003-style-guide.md
+++ b/kb/0003-style-guide.md
@@ -5,7 +5,7 @@ _[back to Index](../index.md)_
 
 **Related articles**: TBD
 
-*Note: This guideline is based on https://github.com/lefticus/cppbestpractices*
+*Note: This guideline is based on [Collaborative Collection of C++ Best Practices](https://github.com/lefticus/cppbestpractices)*
 
 Consistency is the most important aspect of style. The second most important aspect is following a style that the average C++ programmer is used to reading.
 
@@ -17,8 +17,8 @@ This project uses `.clang-format` file that specifies the formatting style. Whil
 
  * This project uses CamelCase
  * Types (classes, structs, enums, etc) start with upper case: `MyClass`.
- * Functions and variables start with lower case: `myMethod`.
- * Constants are all upper case: `const double PI=3.14159265358979323;`.
+ * Functions and variables start with upper case to respect legacy code: `MyMethod()`.
+ * Constants have `k` prefix: `const double kPi=3.14159265358979323;`.
  * Enum values start with upper case (`enum class MyEnum {VariantOne, VariantTwo}`) 
  * Private data must be with a `m_` prefix to distinguish it from public data. `m_` stands for "member" data.
 
@@ -26,7 +26,7 @@ This project uses `.clang-format` file that specifies the formatting style. Whil
 
 If you do, you risk colliding with names reserved for compiler and standard library implementation use:
 
-http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier
+<http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier>
 
 
 ## Well-Formed Example
@@ -39,7 +39,7 @@ class MyClass
     {
     }
 
-    int getData() const
+    int GetData() const
     {
         return m_data;
     }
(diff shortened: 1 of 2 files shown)


kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+ * Functions and variables start with lower case: `myMethod`.
+ * Constants are all upper case: `const double PI=3.14159265358979323;`.
+ * Enum values start with upper case (`enum class MyEnum {VariantOne, VariantTwo}`) 
+ * Private data must be with a `m_` prefix to distinguish it from public data. `m_` stands for "member" data.
+
+## Don't Name Anything Starting With `_`
+
+If you do, you risk colliding with names reserved for compiler and standard library implementation use:
+
+http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier
```
np-c0mp1ete: I think it's better to create a clickable link.

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+
+C++ allows for arbitrary-length identifier names, so there's no reason to be terse when naming things. Use descriptive names, and be consistent in the style.
+
+This project uses `.clang-format` file that specifies the formatting style. While this cannot help with naming, it is particularly important to maintain a consistent style.
+
+## Naming Conventions
+
+ * This project uses CamelCase
+ * Types (classes, structs, enums, etc) start with upper case: `MyClass`.
+ * Functions and variables start with lower case: `myMethod`.
```
np-c0mp1ete: Currently, most of the codebase uses `MyMethod` style. I guess we should stick to it just to not introduce a major inconsistency between the code and the guide.

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+It pollutes the namespace and it may lead to name collisions in the future.
+Writing `using namespace` in an implementation file is fine though.
+
+
+## Include Guards
+
+Use the `#pragma once` directive instead include guards which is quasi-standard across many compilers.
+It's short and makes the intent clear.
+
+## Use "" for Including Local Files
```
np-c0mp1ete: We had a discussion about it in Discord some time ago. We agreed to use "" syntax only for relative paths and <> for files from other libs.
https://discord.com/channels/353218868896071692/820042176125730816/829331810046967848

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+};
+```
+
+Since a const member variable cannot be assigned a new value, such a class may not have a meaningful copy assignment operator.
+
+## Always Use Namespaces
+
+There is almost never a reason to declare an identifier in the global namespace. Instead, functions and classes should exist in an appropriately named namespace or in a class inside of a namespace. Identifiers which are placed in the global namespace risk conflicting with identifiers from other libraries (mostly C, which doesn't have namespaces).
+
+This project uses `storm` namespace
```
np-c0mp1ete: A dot at the end of a sentence is missing. Sorry for being a grammar nazi :)

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+const auto diff = s1 - s2; // diff underflows to a very large number
+```
+
+## Use .hpp and .cpp for Your File Extensions
+
+Ultimately this is a matter of preference, but .hpp and .cpp are widely recognized by various editors and tools. So the choice is pragmatic. Specifically, Visual Studio only automatically recognizes .cpp and .cxx for C++ files, and Vim doesn't necessarily recognize .cc as a C++ file.
+
+One particularly large project ([OpenStudio](https://github.com/NREL/OpenStudio)) uses .hpp and .cpp for user-generated files and .hxx and .cxx for tool-generated files. Both are well recognized and having the distinction is helpful.
+
+## Never Mix Tabs and Spaces
```
np-c0mp1ete: Mention explicitly that we use spaces everywhere?

espkk:
Thanks for the PR!
There are some points I disagree. They should be discussed.
I will have some time to review this in a few days, please wait for the feedback

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+
+Know and understand the existing C++ standard algorithms and put them to use.
+
+ * See [cppreference](https://en.cppreference.com/w/cpp/algorithm)
+ * Watch [C++ Seasoning](https://www.youtube.com/watch?v=qH6sSOr-yk8)
+ 
+Consider a call to `[]` as a potential code smell, indicating that an algorithm was not used where it could have been.
+
+
+## Never Use `assert` With Side Effects
```
np-c0mp1ete: Seems to be a duplication of `Never Put Code with Side Effects Inside an assert()`

np-c0mp1ete (commented):
Thanks a lot for the PR!
IMO, we should try to make it shorter. I'd rather have less guidelines that are followed than a lot that no one reads.
Most of the points are generic, so I'd remove them keeping a link to cppbestpractices and focus on things that are more relevant to our specific project:
- Naming Conventions
- Use "" for Including Local Files
- Always Use Namespaces
- Use .hpp and .cpp for Your File Extensions
- Never Mix Tabs and Spaces
- Avoid Raw Memory Access
- Use `std::array` or `std::vector` Instead of C-style Arrays
- Use Exceptions
- Use C++-style cast instead of C-style cast
- Avoid Compiler Macros
- Use The Standard Library
- Avoid Global Data
- Avoid Excess Exceptions

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+C++ allows for arbitrary-length identifier names, so there's no reason to be terse when naming things. Use descriptive names, and be consistent in the style.
+
+This project uses `.clang-format` file that specifies the formatting style. While this cannot help with naming, it is particularly important to maintain a consistent style.
+
+## Naming Conventions
+
+ * This project uses CamelCase
+ * Types (classes, structs, enums, etc) start with upper case: `MyClass`.
+ * Functions and variables start with lower case: `myMethod`.
+ * Constants are all upper case: `const double PI=3.14159265358979323;`.
```
espkk: I'd vote for google style kConstantName. UPPERCASE is for macros

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+ * Enum values start with upper case (`enum class MyEnum {VariantOne, VariantTwo}`) 
+ * Private data must be with a `m_` prefix to distinguish it from public data. `m_` stands for "member" data.
```
espkk: I prefer postfix '_'
We shall discuss that with others

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+
+In C++11 you can assign default values to each member (using `=` or using `{}`).
+
+### Assigning default values with =
+
+```cpp
+// ... //
+  private:
+    int m_value = 0; // allowed
+    unsigned m_value_2 = -1; // narrowing from signed to unsigned allowed
```
espkk: I would not allow this. Every SCA will alarm that

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+}
+```
+
+## Consider Avoiding Boolean Parameters
+
+They do not provide any additional meaning while reading the code. You can either create a separate function that has a more meaningful name, or pass an enumeration that makes the meaning more clear.
+
+See http://mortoray.com/2015/06/15/get-rid-of-those-boolean-function-parameters/ for more information.
+
+## Avoid Raw Loops
```
espkk: This is a controversial point. In general, simple loops are way more expressive. especially range-based loops

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+[[maybe_unused]] const auto success = setValue(something);
+assert(success);
+```
+
+The `assert()` will be removed in release builds which will prevent the `setValue` call from ever happening.
+
+So while the second version is uglier, the first version is simply not correct.
+
+
+## Properly Utilize 'override' and 'final'
```
espkk: I'd add "always use override if you are actually overriding something"

kb/0003-style-guide.md
```diff
@@ -0,0 +1,1004 @@
...
+
+See also [related safety discussion](#consider-return-by-value-for-mutable-data-const--for-immutable) about `const &` return values
+
+# Considering Performance
+
+## Build Time
+
+
+
+### Forward Declare When Possible
```
espkk: I know this approach but I'm strongly against it.

espkk (commented):
These are the points I noticed. Overall it is fine, but I agree with @np-c0mp1ete: ideally, we want to keep the guidelines brief as possible, and maybe leave some rationale or specific case links

kb/0003-style-guide.md
np-c0mp1ete: Personally, I'm for `m_`

kb/0003-style-guide.md
```diff
@@ -0,0 +1,210 @@
...
+// makes code more difficult to package and distribute.
+#include <string>
+#include <storm/OtherLibHeader.hpp>
+#include <MyLibHeader.hpp>
+
+
+// Good Idea
+// Requires no extra params and notifies the user that the file
+// is a local file.
+#include <string>
```
np-c0mp1ete, line 64: maybe leave only the desired option and remove both bad ones?

kb/0003-style-guide.md
```diff
@@ -0,0 +1,210 @@
...
+
+There is almost never a reason to declare an identifier in the global namespace. Instead, functions and classes should exist in an appropriately named namespace or in a class inside of a namespace. Identifiers which are placed in the global namespace risk conflicting with identifiers from other libraries (mostly C, which doesn't have namespaces).
+
+This project uses `storm` namespace.
+
+## Use .hpp and .cpp for Your File Extensions
+
+Ultimately this is a matter of preference, but .hpp and .cpp are widely recognized by various editors and tools. So the choice is pragmatic. Specifically, Visual Studio only automatically recognizes .cpp and .cxx for C++ files, and Vim doesn't necessarily recognize .cc as a C++ file.
+
+One particularly large project ([OpenStudio](https://github.com/NREL/OpenStudio)) uses .hpp and .cpp for user-generated files and .hxx and .cxx for tool-generated files. Both are well recognized and having the distinction is helpful.
```
np-c0mp1ete: this paragraph is not very useful for us and can be removed.

np-c0mp1ete (commented):
Thanks for the adjustments! I think the current version is a good starting point and can be merged once we decide between `m_` prefix and `_` postfix.

kb/0003-style-guide.md
np-c0mp1ete: we agreed to use postfix '_'

np-c0mp1ete:
@ObKo thanks a lot for all your efforts! I find the Style Guide ready to be merged.
Could you address the single unresolved discussion about prefix `m_` vs postfix `_` and rebase your branch?

ObKo:
Should I squash commits?

np-c0mp1ete:
> Should I squash commits?

no, it's fine, better to not lose the history of changes.

espkk:
This will work for now. Thank you!

PR #51, opened 2021-04-11
INIFILE interface cleanup

vorot93:
* Remove several unused methods
* Replace stateful iteration methods in public API with `ForEach` for cleaner code and potential for `const`ness

### DIFF
--- a/src/apps/ENGINE/file_service.cpp
+++ b/src/apps/ENGINE/file_service.cpp
@@ -468,8 +468,3 @@ bool INIFILE_T::GetSectionNameNext(char *section_name_buffer, long buffer_size)
 {
     return ifs_PTR->GetSectionNameNext(section_name_buffer, buffer_size);
 }
-
-void INIFILE_T::Flush()
-{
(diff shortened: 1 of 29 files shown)


src/apps/ENGINE/file_service.cpp
```diff
@@ -386,120 +386,156 @@ void INIFILE_T::WriteDouble(const char *section_name, const char *key_name, doub
...
+    SEARCH_DATA s;
+
+    auto n = 0;
+    char buffer[2048];
+    while (true)
+    {
+        bool found;
+        if (n == 0)
+        {
+            found = ifs_PTR->ReadString(&s, section_name, key_name, buffer, 2048, "");
```
np-c0mp1ete, line 415: it seems this can be moved out of the loop body. Same applies for the similar code in the other `ForEachString` overload.

np-c0mp1ete (commented):
The code looks fine for me. Have you tested that TEHO still works after the changes? Being able to spawn in a location after starting a new game should be enough to verify the correctness.

PR #52, opened 2021-04-13
Remove exception wrapping

vorot93:
Interferes with debugging, duplicates what OS itself should be doing


espkk:
Thanks for the PR, but we really need these handlers.
I have an idea how to fix that properly. I created an issue https://github.com/storm-devs/storm-engine/issues/55

PR #53, opened 2021-04-13
Fix crash on no SAVE folder

vorot93:


src/apps/ENGINE/file_service.cpp
```diff
@@ -129,7 +129,14 @@ std::vector<std::filesystem::path> FILE_SERVICE::_GetFsPathsByMask(const char *s
     }
 
     std::filesystem::path curPath;
-    for (auto &dirEntry : std::filesystem::directory_iterator(srcPath))
+    std::error_code ec;
+    auto it = std::filesystem::directory_iterator(srcPath, ec);
+    if (ec)
+    {
+        return result;
```
np-c0mp1ete, line 140: Should we also print an error in the log? `core.tracelog->error()`

espkk: I think it should be a warning (not an error). @vorot93 Could you please add logging before return?

src/apps/ENGINE/file_service.cpp
```diff
@@ -133,6 +133,10 @@ std::vector<std::filesystem::path> FILE_SERVICE::_GetFsPathsByMask(const char *s
     auto it = std::filesystem::directory_iterator(srcPath, ec);
     if (ec)
     {
+        if (core.tracelog)
+        {
+            core.tracelog->warn("Failed to open save folder: {}", ec.message());
```
espkk, line 138: thanks for the change
but generally, GetFsPathsByMask can be used not only for querying the save folder
it would be better to look like "failed to get directory_iterator for {}", srcPath

espkk (approved):

PR #54, opened 2021-04-14
Start work on Functions Reference

Yakvi:
Related Task: https://github.com/storm-devs/storm-engine/projects/4#card-58081815

The documentation on the scripting functions is sparse and incomplete, so I dug into the compiler code to have a complete overview of the functions. Turns out, some are never used in the current game files!

In any case, in the following days I'm planning to add descriptions, determine return types etc. as needed. 

Updates: 
* Add general info on the compiler operation in 0002 (will probably expand in the future)
* Fetch the lists of functions in existing documentation
* Fetch the lists of internal functions from `internal_functions.cpp`
* Settle on a formatting style ([preview](https://yakvi.github.io/storm-engine/kb/0003-scripting-functions.html))


kb/0002-scripting-overview.md
```diff
@@ -40,6 +40,13 @@ The following list is sorted alphabetically and is based on the latest build of
 * `program\weather\` - weather system definitions and usage.
 * `program\worldMap\` - everything world map related.
 
+## Runtime
+
+The goes through the following phases during script execution:
```
np-c0mp1ete: > The goes

looks wrong

kb/0003-scripting-functions.md
```diff
@@ -5,7 +5,711 @@ _[back to Index](../index.md)_
...
+    * Because the value affects the whole game, make sure you don't forget to return the time scale back to normal! 
+    ``` C++
+    syntax:
+        void SetTimeScale(int value);
+        void SetTimeScale(float value);
+    usage: 
+        SetTimeScale(0.0);
+    ```
+
+NOTE(yakvi): Below is work-in-progress
```
np-c0mp1ete, line 60: You want to leave it as a reminder for yourself?

kb/0003-scripting-functions.md
```diff
@@ -5,7 +5,711 @@ _[back to Index](../index.md)_
...
+        string GetAttributeValue(aref attribute);
+    ```
+
+* **Vartype**
+    * **Compiler Token**: `FUNC_VARTYPE`
+    ``` C++
+    syntax:
+        string Vartype(aref attribute);
+
+    /* TODO(yakvi): actually confirm it's an aref input, from the source looks like it
```
np-c0mp1ete: @mitrokosta, could you confirm it?

kb/0003-scripting-functions.md
```diff
@@ -5,7 +5,711 @@ _[back to Index](../index.md)_
...
+    syntax:
+        void Breakpoint();
+    ```
+
+* **Pow**
+    **Compiler Token**: `FUNC_POW`
+    ``` C++
+    syntax:
+        float Pow(float exponent, float base);
+        // TODO(yakvi): base/exponent inverted??? to verify
```
np-c0mp1ete, line 483: @mitrokosta, could you help here as well?

kb/0003-scripting-functions.md
Yakvi: Yes, and for the readers to indicate that below is not complete.

kb/0002-scripting-overview.md
Yakvi: Reworked the phrase.

kb/0003-scripting-functions.md
mitrokosta: Vartype was never used in scripts, therefore I can't confirm or dispute it based on script usage. From the source to me it looks like it's actually ref input, not aref.

kb/0003-scripting-functions.md
np-c0mp1ete: from looking at the [code](https://github.com/storm-devs/storm-engine/blob/develop/src/apps/ENGINE/internal_functions.cpp#L1256) and comparing with the signature of [pow](https://www.cplusplus.com/reference/cmath/pow/) it seems to be the case indeed. Unless the script function signature is `float Pow(float base, float exponent);`.
@mitrokosta what do you think?

Yakvi: I'll leave this TODO there for now. It's still a stub, and I'll be able to verify it by the time I get to the function.

mitrokosta: Sorry for taking so long to answer. It is not inverted, it is actually float pow(float base, float exponent). I verified it in scripts: Log_Info(""+pow(10.0, 3.0)); yields 1000. Because of the nature of stack data structure all function arguments are inverted **in engine**. This means that a script function void foo(x, y, z) will have it arguments extracted like this: pV->Get(z); pV->Get(y); pV->Get(x);

Yakvi: That's incredibly useful info that I will use for the multiparam functions! I was suspecting something like this, and having a confirmation is absolutely perfect.

Issue #55, opened 2021-04-14
Persist backtrace on SEH exception


espkk:
Backtrace shall persist when a debugger is attached after SEH exception was thrown

PR #56, opened 2021-04-14
Rework exception handling

espkk:
- Rework exception handling
- Improve core dumps (add exception info on assert())
- Manage the lifetime of CDebug

Addresses #55


espkk:
@np-c0mp1ete wtf is that clang-format check hook? it does not accept code formatted with the latest clang-format

np-c0mp1ete:
> @np-c0mp1ete wtf is that clang-format check hook? it does not accept code formatted with the latest clang-format

I believe that's because you have to use the one that comes with the latest VS and that is clang-format version 11.0.0.

espkk:
> 
> 
> > @np-c0mp1ete wtf is that clang-format check hook? it does not accept code formatted with the latest clang-format
> 
> I believe that's because you have to use the one that comes with the latest VS and that is clang-format version 11.0.0.

well... idk how to format all the files with internal clang-format. could you help here?

np-c0mp1ete:
> > > @np-c0mp1ete wtf is that clang-format check hook? it does not accept code formatted with the latest clang-format
> > 
> > 
> > I believe that's because you have to use the one that comes with the latest VS and that is clang-format version 11.0.0.
> 
> well... idk how to format all the files with internal clang-format. could you help here?

Try `for /R %f in (*.cpp *.h *.hpp) do clang-format -i "%f"` from VS Command Prompt. This command assumes that you are in the src folder

src/apps/ENGINE/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ project(engine)
 
 find_package(Fmod REQUIRED) # required to set ${Fmod_LIBRARY_PATHS}
 
+if(WIN32)
+    add_definitions("/EHa")
```
np-c0mp1ete, line 6: - Why not `add_compile_options()`?
- Isn't it MSVC-specific? I'd suggest putting it in https://github.com/storm-devs/storm-engine/blob/develop/src/CMakeLists.txt next to other compiler-specific options

np-c0mp1ete (changes requested):
Naming conventions in exceptions.hpp don't follow the ones from the Style Guide PR and the rest of the codebase in general.

espkk:
switched to https://github.com/storm-devs/storm-engine/pull/58 solution for now

PR #58, opened 2021-04-14
Add crashpad

### LINKED ISSUE #55: Persist backtrace on SEH exception

espkk:
Just add crashpad finally to get rid of dancing with exceptions.
The only stack trace is uploaded for now. To be tweaked in the future.


espkk:
@np-c0mp1ete could you check what happened to clang build?

np-c0mp1ete:
> @np-c0mp1ete could you check what happened to clang build?

I'm not sure bincrafters support clang builds on Windows. https://github.com/bincrafters/community/pull/1381. Check the list of tests they run on their recipes.

espkk:
> > @np-c0mp1ete could you check what happened to clang build?
> 
> I'm not sure bincrafters support clang builds on Windows. https://github.com/bincrafters/community/pull/1381. Check the list of tests they run on their recipes.

Well, then do we really want to have clang builds on windows?

np-c0mp1ete:
> > > @np-c0mp1ete could you check what happened to clang build?
> > 
> > 
> > I'm not sure bincrafters support clang builds on Windows. [bincrafters/community#1381](https://github.com/bincrafters/community/pull/1381). Check the list of tests they run on their recipes.
> 
> Well, then do we really want to have clang builds on windows?

I'd like to have clang builds in general. It seems there is not much support for clang on windows though. maybe we should indeed drop clang until we are able to compile our code on linux.

espkk:
> > > > @np-c0mp1ete could you check what happened to clang build?
> > > 
> > > 
> > > I'm not sure bincrafters support clang builds on Windows. [bincrafters/community#1381](https://github.com/bincrafters/community/pull/1381). Check the list of tests they run on their recipes.
> > 
> > Well, then do we really want to have clang builds on windows?
> 
> I'd like to have clang builds in general. It seems there is not much support for clang on windows though. maybe we should indeed drop clang until we are able to compile our code on linux.

Maybe there is a way to add msvc-clang build instead? It is literally clang that is much more supported on windows I guess
Anyway, I'm afraid that the current configuration will prevent us from integrating some cool things
We probably should discuss that

CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 # Setup third-party dependencies
 include(conan)
+conan_add_remote(NAME bincrafters
+    URL https://api.bintray.com/conan/bincrafters/public-conan
```
np-c0mp1ete: that's a deprecated url. check the [docs](https://bincrafters.readthedocs.io/en/latest/using_packages.html#adding-the-bincrafters-repository-as-a-conan-remote) and how @q4a [integrated sdl2](https://github.com/q4a/storm-engine/commit/0a78c174129c2b265324e5b155766b80bb3076cd#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR10)

src/apps/ENGINE/crashpad.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "vfile_service.h"
+
+#include <client/crash_report_database.h>
+#include <client/crashpad_client.h>
+#include <client/settings.h>
+
+namespace storm::except
+{
+
+bool startCrashHandler()
```
np-c0mp1ete: Most of the codebase uses CamelCase style. Probably we should stick to it.

np-c0mp1ete (changes requested):

CMakeLists.txt
espkk: Not on binary? :O

src/apps/ENGINE/crashpad.cpp
espkk: Our coding style describes exactly this naming...

np-c0mp1ete: I know, I left a comment there in this regard.

CMakeLists.txt
np-c0mp1ete: what do you mean?

espkk: T9 :( I meant bintray. I believe it's not deprecated but rather a release repo link...

espkk: Ok I see it's getting closed soon

src/apps/ENGINE/crashpad.cpp
espkk: as discussed, let's leave it as is right now as global reformat incoming

src/CMakeLists.txt
```diff
@@ -7,6 +7,10 @@ if (STORM_ENABLE_STEAM)
 endif()
 
 if (MSVC)
+    # Output to the same directory regardless of [CONFIG]
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
```
np-c0mp1ete: I tested it on my PC and I don't have this issue with subdirectories inside `bin`.

np-c0mp1ete (commented):
Thanks for the PR! How can I access the sentry dashboard?

src/CMakeLists.txt
espkk: as discussed, this problem appears when generating sln rather than using CMakeLists directly

np-c0mp1ete (approved):

PR #63, opened 2021-04-21
Change Windows-specific ENGINE timer to std::chrono.

szymonj99:
I've replaced the Windows-only API's with cross-platform code for the ENGINE timer only.
Compiled and tested locally, with no issues.
Performance seems exactly the same.

### DIFF
--- a/src/apps/ENGINE/timer.h
+++ b/src/apps/ENGINE/timer.h
@@ -76,7 +76,7 @@ class TIMER
             return FixedDeltaValue;
         }
         return Delta_Time;
-    };
+    }
 
     uint32_t GetDeltaTime()
     {
--- a/src/apps/ENGINE/timer.h
+++ b/src/apps/ENGINE/timer.h
@@ -1,112 +1,102 @@
 #pragma once
 
-#include <Windows.h>
-
-#define USE_HIGH_FREQUENCY
-#define FILTER_SIZE 16
+#include <chrono>
 #include <cstdint>
 
 class TIMER
 {
+    using Clock = std::chrono::high_resolution_clock;
+    using Milliseconds = std::chrono::milliseconds;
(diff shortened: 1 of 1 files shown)


src/apps/ENGINE/timer.h
```diff
@@ -1,112 +1,102 @@
...
+        // if (FixedDelta)
+        //{
+        //    return FixedDeltaValue;
+        //}
+        // return Delta_Time;
+
+        // Branchless approach:
+        // If FixedDelta is true, this evaluates to return the FixedDeltaValue.
+        // If not, Delta_Time is returned.
+        return (uint32_t)((FixedDelta * FixedDeltaValue) + (!FixedDelta * Delta_Time));
```
Hammie: I feel like this is an unnecessary micro-optimization. Is there a substantial reason why we would need to avoid branching here?

src/apps/ENGINE/timer.h
```diff
@@ -1,112 +1,102 @@
...
-        rDelta_Time = Delta_Time;
-        fps = 0;
+        Current = Previous = Clock::now();
+    }
+
+    void RefreshFPS()
+    {
+        // Modulo operator is not necessary as we have limited maximum frame delta, so fps_time above 2000 milliseconds
+        // are not possible.
+        fps_time -= (uint32_t)Milliseconds{1000}.count();
```
Hammie: Consider making the `1000` a static const (something like `FpsUpdateInterval`) so it's extra clear what the number represents.

If you'd then also make `kMaxDelta` static, you could add a `static_assert(kMaxDelta < FpsUpdateInterval, "Max delta time should be less than fps update interval");` to verify the no-modulus requirement. Which can be very useful should the numbers ever be tweaked.

np-c0mp1ete (dismissed):
Thanks a lot for the PR!

np-c0mp1ete (approved):

PR #64, opened 2021-04-23
Removed unnecessary ENGINE timer rDelta_Time assignment

szymonj99:
A very minor change to stop rDetla_Time from being assigned twice to the same value.


np-c0mp1ete (approved):
Thanks!

PR #71, opened 2021-05-01
Restore default behaviour for script function SendMessage

mitrokosta:
SendMessage's default behavior when first argument was not a valid entity was to return 0, according to [this](https://github.com/storm-devs/storm-engine/blob/76e8eaf9cde9ed95dc5b65e34e57a3613d387242/ENGINE/Sources/internal_functions.cpp#L1552) line. This was changed in thunderstorm. Some code in TEHO stopped functioning because of that, for example F2 interface wouldn't open on worldmap. While this is something that can and should be fixed in scripts, I'm pretty sure COAS relied on SendMessage returning 0 too, so this behavior should be reverted for compatibility reasons.

### DIFF
--- a/src/apps/ENGINE/internal_functions.cpp
+++ b/src/apps/ENGINE/internal_functions.cpp
@@ -1845,7 +1845,7 @@ DATA *COMPILER::BC_CallIntFunction(uint32_t func_code, DATA *&pVResult, uint32_t
 
         CreateMessage(&ms, s_off, 1);
 
-        uint64_t mresult = -1;
+        uint64_t mresult = 0;
         pE = EntityManager::GetEntityPointer(ent);
         if (pE)
         {


espkk:
What exactly does it return? if it returns entity_id and we want to use 0 there we have to make sure that entity_id==0 would never been created

mitrokosta:
Script function SendMessage returns whatever the entity's ProcessMessage returns, which is could be any uint64_t. This uint64_t is definitely not entity_id. It actually does not make sense to return a ptr to scripts, because the scripts expect an int (long). I'll make another commit that will make FUNC_SEND_MESSAGE return long.

espkk:
I see. But I remember that there was actually some particular reason why I had changed it to -1. This was related to either entity management change or PTR used in code (normally shall be replaced with a handle)
We can merge it as is, but we also shall keep in mind this might cause weird issues

espkk (approved):

mitrokosta:
I see what you are referring to: [link](https://github.com/storm-devs/sd-teho-public/blob/e7d161eb1d4da25a73a517ff6d8373d8acbb6609/program/interface/save_load.c#L962). I'll try then to replace this ptr with a texture handle.

espkk:
> 
> 
> I see what you are referring to: [link](https://github.com/storm-devs/sd-teho-public/blob/e7d161eb1d4da25a73a517ff6d8373d8acbb6609/program/interface/save_load.c#L962). I'll try then to replace this ptr with a texture handle.

Yeah, exactly! ptr in scripts is a pseudo-type I introduced as a workaround there.
If we will use int32_t handle instead it will go away as well.
Feel free to contact me if you will need any help implementing this.

espkk (changes requested):
discussed above

src/apps/ENGINE/internal_functions.cpp
```diff
@@ -1845,7 +1845,7 @@ DATA *COMPILER::BC_CallIntFunction(uint32_t func_code, DATA *&pVResult, uint32_t
 
         CreateMessage(&ms, s_off, 1);
 
-        uint64_t mresult = -1;
+        uint64_t mresult = 0;
```
np-c0mp1ete, line 1848: btw, I've just remembered that it was me who initialized `mresult` with -1. It wasn't initialized before and the address sanitizer caught an uninitialized variable access here. Check the commit ec6dee6 - Fix reading uninitialized variable

espkk: Well...well, maybe it's not related then and safe to merge? But why that change causing issues if it was uninitialized before?

np-c0mp1ete: I think the engine wasn't properly tested after merging thunderstorm (the initialization was removed in 9fdf420 - Build engine executable)

espkk (dismissed):
Indeed, that my be accidentally removed when making the code ISO compliant. If the return code is not related to PTR pseudo-type feel free to rebase and merge this commit

mitrokosta:
I guess it can be merged now because it does not break anything and it fixes not being able to open F2 interface on global map. But, SendMessage should return a long, not ptr, because all scripts except for save_load.c expect SendMessage to return an int. This means some additional work needs to be done - make save_load request texture handles, not pointers. This work can be done in another PR.

Rebasing and force pushing dismissed your review...

espkk (approved):
concluded, thanks!

PR #73, opened 2021-05-07
Restore the ability to get random texture handles from scripts

mitrokosta:
Fixes users being unable to change sail textures.
My version of the fix requires some script changes, that I will soon upload to sd-teho-public


espkk (approved):
Looks nice, thanks!

PR #74, opened 2021-05-11
Fix crash caused by TEHO's composite masts

mitrokosta:
For reference, it used to be like [this](https://github.com/storm-devs/storm-engine/blob/76e8eaf9cde9ed95dc5b65e34e57a3613d387242/Ship/ship.cpp#L956), but was overwritten at some point. Was causing crashes in naval battles, because the same object (node) was destroyed twice.


espkk (approved):
I assume this should be compatiable with CoAS, then it seems fine, thanks!

PR #75, opened 2021-05-15
Replace custom hash table with std::unordered_map

mitrokosta:
I replaced the "HASHLINE" in variables and function tables with std::unordered_map.

I rewrote both tables from scratch and I'm not sure if my naming code style was correct. I tried to conform to Google code style, because it was the most voted one.

Requires a small script change: in [this](https://github.com/storm-devs/sd-teho-public/blob/e7d161eb1d4da25a73a517ff6d8373d8acbb6609/program/characters/LSC_Q2Utilite.c#L1352) line a variable with an empty name is declared which normally should be a compile error. Just need to remove the comma and everything will compile correctly.

### DIFF
--- a/src/apps/ENGINE/s_functab.cpp
+++ b/src/apps/ENGINE/s_functab.cpp
@@ -205,20 +205,26 @@ bool FuncTable::GetVar(LocalVarInfo &lvi, size_t func_index, size_t var_index) c
     return true;
 }
 
-void FuncTable::AddTime(size_t func_index, uint64_t time)
+bool FuncTable::AddTime(size_t func_index, uint64_t time)
 {
-    if (func_index < funcs_.size())
+    if (func_index >= funcs_.size())
     {
-        funcs_[func_index].usage_time += time;
+        return false;
     }
+
+    funcs_[func_index].usage_time += time;
+    return true;
 }
 
-void FuncTable::AddCall(size_t func_index)
+bool FuncTable::AddCall(size_t func_index)
 {
-    if (func_index < funcs_.size())
+    if (func_index >= funcs_.size())
     {
-        ++funcs_[func_index].number_of_calls;
+        return false;
     }
+
+    ++funcs_[func_index].number_of_calls;
+    return true;
 }
 
 void FuncTable::Release()
--- a/src/apps/ENGINE/s_functab.cpp
+++ b/src/apps/ENGINE/s_functab.cpp
@@ -1,19 +1,9 @@
 #include "s_functab.h"
 
-FuncInfo::FuncInfo() : 
-  name(),
-  local_vars(),
-  segment_id(INVALID_SEGMENT_ID),
-  offset(INVALID_FUNC_OFFSET),
-  stack_offset(),
-  arguments(),
-  return_type(TVOID),
-  decl_file_name(),
-  decl_line(),
-  usage_time(),
-  number_of_calls(),
-  imported_func(),
-  extern_arguments()
+FuncInfo::FuncInfo()
+    : name(), local_vars(), segment_id(INVALID_SEGMENT_ID), offset(INVALID_FUNC_OFFSET), stack_offset(), arguments(),
+      return_type(TVOID), decl_file_name(), decl_line(), usage_time(), number_of_calls(), imported_func(),
+      extern_arguments()
 {
 }
 
@@ -78,7 +68,7 @@ bool FuncTable::GetFunc(FuncInfo &fi, size_t func_index) const
         {
             return false;
         }
-        
+
         fi = funcs_[func_index]; // copy func info
         return true;
     }
--- a/src/apps/ENGINE/s_functab.cpp
+++ b/src/apps/ENGINE/s_functab.cpp
@@ -33,16 +33,16 @@ size_t FuncTable::AddFunc(const FuncInfo &fi)
 
     auto result = hash_table_.try_emplace(fi.name, funcs_.size()); // find or create index
     auto func_index = result.first->second;
-    bool isNew = result.second;
+    bool is_new = result.second;
 
-    if (isNew) // newly created function, add to table
+    if (is_new) // newly created function, add to table
     {
         funcs_.push_back(fi);
     }
 
     auto &func = funcs_[func_index];
 
-    if (!isNew)
+    if (!is_new)
     {
         if (func.offset != INVALID_FUNC_OFFSET) // function already loaded
         {
--- a/src/apps/ENGINE/s_functab.cpp
+++ b/src/apps/ENGINE/s_functab.cpp
@@ -1,387 +1,238 @@
 #include "s_functab.h"
 
-/*DWORD _testfunc(VS_STACK * ps)
+FuncInfo::FuncInfo() : 
+  name(),
+  local_vars(),
+  segment_id(INVALID_SEGMENT_ID),
+  offset(INVALID_FUNC_OFFSET),
+  stack_offset(),
+  arguments(),
+  return_type(TVOID),
+  decl_file_name(),
+  decl_line(),
+  usage_time(),
+  number_of_calls(),
+  imported_func(),
+  extern_arguments()
 {
-    return 7;
-}*/
-
-#define MAKEHASHINDEX(x) (x & 0xff)
-
-S_FUNCTAB::S_FUNCTAB()
-{
-    Buffer_size = 0;
-    Func_num = 0;
-    for (uint32_t n = 0; n < HASHT_SIZE; n++)
-    {
-        HashLine[n].nNumElements = 0;
-    }
 }
 
-S_FUNCTAB::~S_FUNCTAB()
+FuncTable::~FuncTable()
 {
     Release();
 }
 
-void S_FUNCTAB::Release()
+size_t FuncTable::FindFunc(const std::string &func_name) const
 {
-    uint32_t n;
-    for (n = 0; n < Func_num; n++)
+    auto result = hash_table_.find(func_name);
+
+    if (result == hash_table_.end())
     {
-        delete pTable[n].name;
-        for (uint32_t i = 0; i < pTable[n].var_num; i++)
-        {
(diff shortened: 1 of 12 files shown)


espkk:
Thanks for the PR! That is something that had to be done for sure.

The only thing I'm a bit concerned about is that boost/std::unordered_list is relatively slow in comparison to some custom (e.g. google's) hashtable implementions. I did some measurements a while ago specifically for SE but it seems that I lost everything :(

But anyway, switching to the std from self-made hashtable already should be winning in terms of performance and maintainability.

I will look though it soon

src/libs/Common_h/strutils.h
```diff
@@ -26,3 +27,32 @@ inline void rtrim(std::string &str)
...
+        std::string lower_copy = key;
+        std::transform(lower_copy.begin(), lower_copy.end(), lower_copy.begin(), ::tolower);
+        return inner_hasher_(lower_copy);
+    }
+
+  private:
+    std::hash<std::string> inner_hasher_;
+};
+
+class CaseInsensitiveStringComparator
```
np-c0mp1ete: there is also a function for case-insensitive string comparison from @Hammie in https://github.com/storm-devs/storm-engine/blob/develop/src/libs/util/include/storm/string_compare.hpp

could it be reused/unified? if not, maybe at least put them in a single file?

src/apps/ENGINE/s_vartab.cpp
```diff
@@ -1,231 +1,117 @@
...
-    uint32_t n;
-    uint32_t hash;
-    if(var_name == 0) return INVALID_VAR_CODE;
-    hash = MakeHashValue(var_name);
-    for(n=0;n<Var_num;n++)
+    if (var_index >= vars_.size())
     {
-      if(pTable[n].hash == hash) //return n;
-      if(_stricmp(pTable[n].name,var_name)==0) return n;
+        return;
```
np-c0mp1ete: maybe log a warning here?

src/apps/ENGINE/s_vartab.cpp
```diff
@@ -1,231 +1,117 @@
...
-bool S_VARTAB::ArraySizeChanged(uint32_t nIndex, uint32_t nNewSize)
-{
-    if (nIndex >= Var_num)
-        return false;
-    if (!pTable[nIndex].pDClass->IsArray())
+    if (!vars_[var_index].value->IsArray())
     {
-        // trace("ERROR: invalid variable type for 'ChangeArraySize(?)' function");
-        return false;
+        return;
```
np-c0mp1ete: log a warning? or maybe return a bool and log a warning on the caller site if result != true?

src/apps/ENGINE/s_vartab.cpp
```diff
@@ -1,231 +1,117 @@
...
+    vi = vars_[var_index]; // copy var info
+    return true;
 }
 
-uint32_t S_VARTAB::MakeHashValue(const char *string)
+void VarTable::InvalidateBySegmentID(uint32_t segment_id)
 {
-    uint32_t hval = 0;
-    while (*string != 0)
+    for (auto &vi : vars_)
```
np-c0mp1ete, line 94: would be nice to handle a case when a given segment_id is not found

src/apps/ENGINE/s_functab.cpp
```diff
@@ -1,387 +1,228 @@
...
-        pTable[n].decl_file_name = nullptr;
+        return false;
     }
+
+    fi = funcs_[func_index]; // copy func info
+    return true;
 }
 
-void S_FUNCTAB::InvalidateFunction(uint32_t nFuncHandle)
+void FuncTable::InvalidateBySegmentID(uint32_t segment_id)
```
np-c0mp1ete, line 96: handle a case when `segment_id` is not found?

src/apps/ENGINE/s_functab.cpp
```diff
@@ -1,387 +1,228 @@
...
+    return true;
 }
 
-void S_FUNCTAB::AddCall(uint32_t func_code)
+void FuncTable::AddTime(size_t func_index, uint64_t time)
 {
-    if (func_code >= Func_num)
-        return;
-    pTable[func_code].nNumberOfCalls++;
+    if (func_index < funcs_.size())
```
np-c0mp1ete: return an error when `func_index >= funcs_.size()`?

src/apps/ENGINE/s_functab.cpp
```diff
@@ -1,387 +1,228 @@
...
+    }
 }
 
-void S_FUNCTAB::UpdateHashTable(uint32_t code, uint32_t hash, bool in)
+void FuncTable::AddCall(size_t func_index)
 {
-    const auto hash_index = MAKEHASHINDEX(hash);
-
-    for (uint32_t n = 0; n < HashLine[hash_index].nNumElements; n++)
+    if (func_index < funcs_.size())
```
np-c0mp1ete: return an error when `func_index >= funcs_.size()`?

np-c0mp1ete (commented):
thanks for the PR! 
I think it's better to stick to the standard library unless it's found to be a performance bottleneck. Switching to abseil or another lib shouldn't be hard if needed.

src/libs/Common_h/strutils.h
Hammie: It should do the same, and you could also update mine to use `std::equal` in the process ;)

src/apps/ENGINE/s_vartab.h
```diff
@@ -1,83 +1,59 @@
...
-    DATA *pDClass;
-    char *name;
-    uint32_t elements;
+        return elements > 1;
+    }
+
+    std::string name;
     S_TOKEN_TYPE type;
-    bool bArray;
+    std::shared_ptr<DATA> value;
```
Hammie: Is this shared_ptr used anywhere else? Or could it be a unique_ptr?

mitrokosta: With the current interface of VarTable it can't be replaced with a unique_ptr. VarTable::GetVar and VarTable::GetVarX require to make a copy of a VarInfo.

espkk: I have not had a chance to review the code yet, but I believe we definitely shall try to avoid using std::shared_ptr there
It would be a significant drop in performance due to a control block
I will try to check this soon and propose a solution

mitrokosta: I rewrote the code, so it uses a const pointer to VarInfo instead of a copy. This allowed me to get rid of shared_ptr. Let me know if this is a good solution.

src/apps/ENGINE/s_vartab.cpp
mitrokosta: How do you suggest to handle it? It is not an error - a segment may have no variables at all, theoretically (if it only operates with global variables from another segment, for example).

src/apps/ENGINE/s_functab.cpp
mitrokosta: Same as above, some segments may have no functions at all, only variables.

src/apps/ENGINE/s_functab.cpp
mitrokosta: Done.

src/apps/ENGINE/s_functab.cpp
mitrokosta: Done.

src/apps/ENGINE/s_vartab.cpp
mitrokosta: Done.

src/apps/ENGINE/s_vartab.cpp
mitrokosta: Done.

src/libs/Common_h/strutils.h
mitrokosta: I moved my hasher and comparator to this file, also reused @Hammie 's iEquals function. Also I updated iEquals to use std::equal.

src/apps/ENGINE/s_vartab.cpp
np-c0mp1ete: I see. I guess it's fine to leave it as it is then.

np-c0mp1ete (dismissed):
Thanks for the PR! I have no further comments, but I'll let @espkk to review it before merging too.

src/apps/ENGINE/s_functab.h
```diff
@@ -15,79 +16,84 @@
...
 {
+    LocalVarInfo() = default;
+
+    bool IsArray()
+    {
+        return elements > 1;
+    }
+
+    std::string name;
+    size_t hash; // for faster comparison
```
espkk, line 29: can you explain how it is used? I cannot find it's by inspecting the commit...

mitrokosta: It is used in FuncTable::FindVar, [this](https://github.com/storm-devs/storm-engine/blob/2bea837a898bc16d6209318065fd48a2cdabc1bc/src/apps/ENGINE/s_functab.cpp#L178) line. The same thing was done in the old code.

espkk: So it is hashed twice, right? I recall now that this was another point against stock unordered_map
Though it is still fine for now
Everything else is good! I think this can be merged after rebase

mitrokosta: Variable names in FuncTable are not actually hashed twice, because they are not stored in an unordered_map. They are stored in a vector (local_vars) which is stored in FuncInfo which is stored in an unordered_map. In order to find a functions' argument you have to look that function up first, using unordered_map, and then find the argument in local_vars vector, using this argument's hash to speed up comparison. The same thing was done in the old code, the difference is that it used a custom hash table.

espkk: Ah, that makes sense
I believe I tried to invent a container for string interning. I see now this is unchanged here

espkk (approved):

PR #76, opened 2021-05-15
Overhaul logging and crash handling

espkk:
What is done:
- Enable logging by default
- Flush in a separate thread every 5 seconds (configurable in code so far) after core iteration run
- Fixup sentry/crashpad, attach archived logging folder

There is still some related things to tweak. Mostly to make options configurable without rebuilding. But this is good to merge asap and eventually do the rest.


np-c0mp1ete (dismissed):
Thanks for the PR! I don't have any issues with the code. 
It might be nice to provide an abstraction layer for logging (don't call spdlog functions across the codebase directly but rather storm::logging::error() that would call spdlog::error() internally), but probably that's an overengineering since we don't plan to replace spdlog in the foreseeable future.

ugeen4:
I don't see any problems with this code - successfully compiled and tested by me

ugeen4 (dismissed):

espkk:
> Thanks for the PR! I don't have any issues with the code.
> It might be nice to provide an abstraction layer for logging (don't call spdlog functions across the codebase directly but rather storm::logging::error() that would call spdlog::error() internally), but probably that's an overengineering since we don't plan to replace spdlog in the foreseeable future.

Yeah, as we discussed before I'd against introducing abstractions without obvious necessity. Since we have just a few developers it will drastically decrease maintainability for such simple cases (where spdlog::error can be find-and-replaced with storm::error for instance).

NB: please don't merge it so far. I reviewed the code myself and have some concerns now:
1) the code around fopen is likely unsafe. I have to check the pointer for nullptr. Or better write an own sink and get full filenames instead of relying on a ".log" extension (which is only guaranteed if created with getOrCreateLogger helper). Though upcast there will need RTTI which I'm not really glad of.
2) I'm not sure if my synchronization in LoggingService is fully thread-safe. I need to double check it. I would also appreciate if someone else can take a look too...

Hammie:
Is their a way to disable sentry at runtime? For development but also because some people might not want to automatically share crash logs.

espkk:
> 
> 
> Is their a way to disable sentry at runtime? For development but also because some people might not want to automatically share crash logs.

Not yet, but it definitely will be. As well as some logging configuration.
This is to be placed in the "engine.ini" alternative that I plan to have in `GetStashPath()` folder (in TOML)
But this is a task for another PR.

Also, IMO the user shall be asked for sending crashdata (even if no personal data is processed) with a modal dialog. I plan to do this with wxWidgets inside crashhandler. Also, zipping logs proc should be moved to the crashhandler too (ideally, no logic after crash shall be executed in the main process even though SEH allow this)

Hammie (dismissed):

src/libs/diagnostics/src/LifecycleDiagnosticsService.cpp
```diff
@@ -0,0 +1,197 @@
...
+    bool flushRequested_{false};
+    std::atomic_bool terminate_{true};
+
+    void flushAll(bool doSync)
+    {
+        auto getLogPath = [](std::shared_ptr<spdlog::logger> &l) {
+            return (fs::GetLogsPath() / l->name()).replace_extension(".log");
+        };
+
+        spdlog::apply_all([&getLogPath, doSync](std::shared_ptr<spdlog::logger> l) {
```
np-c0mp1ete: `getLogPath` is unused

np-c0mp1ete (dismissed):
Thanks for the PR! The code seems good, haven't run it, though.
I guess, something is still wrong with the source files filter in VS, cause for the `diagnostics` lib it shows both, `Source Files` and `src` groups, but it's not related to the current PR.

src/libs/diagnostics/src/LifecycleDiagnosticsService.cpp
espkk: fixed, thanks

espkk:
> Thanks for the PR! The code seems good, haven't run it, though.
> I guess, something is still wrong with the source files filter in VS, cause for the `diagnostics` lib it shows both, `Source Files` and `src` groups, but it's not related to the current PR.

Is this indeed about VS or VS Code? 'cause I have the following structure after generation:
image

Probably, you need to clean cmake cache + delete solution files and try to reconfigure. 
If the problem still remains, please, let me know

Hammie (approved):
I quickly tested things, and everything seems to be working. Logs are also correctly flushed when the game crashes.

PR #78, opened 2021-05-22
Fix memory leak caused by incorrect delete operator

CaptainBeltrop:
I fixed a few warnings that were shown by MSVS 2019.


espkk (approved):
Thanks for the PR!
It won't be a memory leak though it's definitely a mistake that should be fixed.
I think we have some more of these in the code because before I ejected memory manager every object was deleted like this.

PR #79, opened 2021-05-23
Message refactor (part 1)

Hammie:
Refactored `MESSAGE` and `SCRIPT_MESSAGE` into a single class and reduced the amount of `va_arg` voodoo.

- The message will now contain its own parameters properly, also making it easier to query the format and parameters of a message. This will be very useful for backwards compatibility work and should also allow improving some of the error messages.
- I removed the memory block and struct parameter types because they were not used.
- I changed the String parameter type to return a proper std::string instead of a char*, getting rid of a lot of messy string buffer code.

I'll do more refactoring later, including getting rid of the `va_arg` stuff entirely, but I don't want to cram too much in a single pull request.


np-c0mp1ete (dismissed):
Thanks a lot for the PR, really appreciated! I looked through the changes and have no comments. I'll let @espkk to take a look as well before merging it.

espkk (dismissed):
Great job! I did not run this but the code seems good to me.
I'm wondering if you compared the performance? 
I remember some serious performance issues with std::variant.
However, it was years ago, so hopefully, it was optimized since then...

espkk:
@Hammie, could you please resolve conflicts and rebase

Hammie:
> 
> 
> Great job! I did not run this but the code seems good to me.
> I'm wondering if you compared the performance?
> I remember some serious performance issues with std::variant.
> However, it was years ago, so hopefully, it was optimized since then...

I didn't do any actual measurements but while testing it in CoAS I did not notice any difference. Don't think we actually process enough messages for it to be noticeable. Since we only use the `std::variant` as a more type-safe union here, I think the only slowdown would come from the extra type checks, but I'll be happy to have those.

espkk (approved):

PR #80, opened 2021-05-26
Remove unused PROGRAM class

mitrokosta:
Two less files to worry about


np-c0mp1ete (approved):
Thanks a lot for the cleanup!

PR #82, opened 2021-05-31
Fix capitalization of filenames in CMake configuration

Hammie:
Fixes the capitalization of filenames in the CMake configuration.

Also adds missing includes for OSWindow when compiling with gcc, and disables the windows-specific options for unit tests.


np-c0mp1ete:
> Fixes the capitalization of filenames in the CMake configuration.

Personally, I'd prefer to fix the naming using `git mv` and have a unified style as the result. What do you think about it?

Hammie:
> Personally, I'd prefer to fix the naming using `git mv` and have a unified style as the result. What do you think about it?

Have we settled on a recommended naming convention yet?

Yes, renaming the files would be better, I'm just a little worried about merge conflicts when moving large amounts of files. I also remember renaming file capitalization can be a major PITA on Windows.

For me, it would make more sense to fix all the file names when we have a semi-working linux build, so we can actually test if the cases are correct.

np-c0mp1ete:
> For me, it would make more sense to fix all the file names when we have a semi-working linux build, so we can actually test if the cases are correct.

Okay, agreed, makes sense.

np-c0mp1ete (dismissed):
Thanks for the PR! Please rebase the branch before merging.

P.S. @espkk and me agreed to use rebase instead of merge to update feature branches to make the git history more linear. It's sometimes hard to follow this agreement with occasional new contributors who are not aware of it. Unfortunately, github [doesn't support this option](https://github.com/isaacs/github/issues/1143), so I can't rebase your branch myself. 
BTW, you could also create branches directly in the upstream repo since you have a collaborator role, then I'll be able to rebase them without bothering you.
Anyway, I just wanted to let you know about this agreement since you're a core member of the development team.

Hammie:
> P.S. @espkk and me agreed to use rebase instead of merge to update feature branches to make the git history more linear. It's sometimes hard to follow this agreement with occasional new contributors who are not aware of it. Unfortunately, github [doesn't support this option](https://github.com/isaacs/github/issues/1143), so I can't rebase your branch myself.

I'm not really a fan of that kind of rebasing, but ok, should be done now.

espkk (approved):

PR #86, opened 2021-06-01
Set default sky rendering techniques

Hammie:
Sets default sky rendering techniques in case scripts do not specify custom ones


espkk (approved):
Thanks!

PR #87, opened 2021-06-01
Fix bug when using strcut to produce single character strings

Hammie:
`std::string(pChar[TempLong1], 1)` should have been `std::string(1, pChar[TempLong1])`

But the special treatment of single character strings no longer seems useful, so I just deleted it.


ugeen4 (approved):
Ok, bug fixed !

PR #91, opened 2021-06-06
Add support for old behavior of GetRiggingData

Hammie:
Fixes flags for CoAS

Since this is the first substantial backwards-compatibility change. I welcome any feedback or concerns about the approach taken.

### DIFF
--- a/src/apps/ENGINE/Core.cpp
+++ b/src/apps/ENGINE/Core.cpp
@@ -275,7 +275,7 @@ void CORE::ProcessEngineIniFile()
             throw std::runtime_error("fail to run program");
 
         // Script version test
-        if (m_targetVersion >= storm::ENGINE_VERSION::LATEST)
+        if (targetVersion_ >= storm::ENGINE_VERSION::LATEST)
         {
             long iScriptVersion = 0xFFFFFFFF;
             auto *pVScriptVersion = static_cast<VDATA *>(core.GetScriptVariable("iScriptVersion"));
@@ -908,6 +908,11 @@ void *CORE::GetScriptVariable(const char *pVariableName, uint32_t *pdwVarIndex)
     return real_var->value.get();
 }
 
+storm::ENGINE_VERSION CORE::GetTargetEngineVersion() const noexcept
+{
+    return targetVersion_;
+}
+
 void CORE::Start_CriticalSection()
 {
     EnterCriticalSection(&lock);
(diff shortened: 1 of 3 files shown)


espkk:
Thanks for the PR!
I still feel like TEHO's GetRiggingData should be GetRiggingDataEx with an additional parameter.
In other words, that msgs/events should not be specialized but unified.
Probably it makes sense to discuss this one more time together?

mitrokosta:
I like this approach! Works with all games and does not require script changes. Also I think it is much easier to implement backward compatibility this way, not through additional TEHO-specific events. Gotta check how it works in game though.

Hammie:
The main reason I want to keep it as `GetRiggingData` is to keep the engine as close to a drop-in-replacement as possible. 

I'm hoping that the easier it is for modders to adopt the new engine, the quicker they actually will. And having more people using the engine should also mean more people willing to help us improve it.

I'm also hoping that in the (distant) future we can get rid of much of this backwards-compatibility code, and everyone can just adopt the newer implementation. In this case for example, the new implementation is just better and more flexible than the old one. Not sure how exactly we would handle this, but I would be sad if we ended up with a `GetRiggingDataEx` but no `GetRiggingData`.

espkk:
I agree this sounds like a good solution, and it will surely work at some point.
But consider the following:
You want COAS to use TEHO's GetRiggingData or some other msg/event in terms of upgrade
With "ex" you would have an option to just switch to GetRiggingDataEx
Instead, you have only these options:
1) switch to TEHO compatibility layer. Will require much overhauling. Also, many changes will be done thus this will in fact diverge baseline from many mods based on COAS
2) introduce new msg/event. Makes no sense since your whole point is to avoid that
3) bring your ideas

We can introduce per-msg layer/version ID but this would be even more complex for both modders and us, who works on engine.

I do not say this is not possible with this approach, but I do not see a suitable solution yet.

Anyway, I would not block this commit, feel free to merge it, but please keep in mind the issue described above.
If we'll able to solve it I think we're completely good with that compatibility stuff

Hammie:
I also agree that this approach will leave a rather large gap between game versions, so modders will have a hard time upgrading to the next if they want. You also mentioned all the solutions I could think of right now, and I agree there are currently no ideal solutions.

Though I do believe we should be able to find a very good solution in the future. If not we may end up introducing a GetRiggingDataEx after all.

@ugeen4 Please tell us if you have anything to add, as you were also working on compatibility. If not we should consider merging this.

Hammie:
@espkk Can we merge this?

espkk:
Of course! As I said, I wouldn't block this. It is probably the best solution so far
Please, rebase it so we can merge

espkk (approved):

PR #93, opened 2021-06-12
Fix memory allocation bug

mitrokosta:
Fixed a bug that was causing heap buffer overflows if the requested size exceeded 4096 bytes;

### DIFF
--- a/src/apps/ENGINE/compiler.cpp
+++ b/src/apps/ENGINE/compiler.cpp
@@ -1088,9 +1088,11 @@ void COMPILER::ProcessFrame(uint32_t DeltaTime)
 
 void COMPILER::ResizeBCodeBuffer(SEGMENT_DESC &Segment, uint32_t add_size)
 {
-    if ((Segment.BCode_Program_size + add_size) >= Segment.BCode_Buffer_size)
+    int64_t needed_memory = static_cast<int64_t>(Segment.BCode_Program_size) + add_size - Segment.BCode_Buffer_size;
+    if (needed_memory >= 0)
     {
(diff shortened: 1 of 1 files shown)


src/apps/ENGINE/compiler.cpp
```diff
@@ -1088,9 +1088,11 @@ void COMPILER::ProcessFrame(uint32_t DeltaTime)
 
 void COMPILER::ResizeBCodeBuffer(SEGMENT_DESC &Segment, uint32_t add_size)
 {
-    if ((Segment.BCode_Program_size + add_size) >= Segment.BCode_Buffer_size)
+    int64_t needed_memory = static_cast<int64_t>(Segment.BCode_Program_size) + add_size - Segment.BCode_Buffer_size;
+    if (needed_memory >= 0)
```
np-c0mp1ete, line 1092: shouldn't it be >0? Otherwise, `needed_memory - 1` would result in a wrong number of blocks.

np-c0mp1ete (changes requested):

src/apps/ENGINE/compiler.cpp
mitrokosta: if needed_memory == 0, then needed_memory - 1 == -1, (needed_memory - 1) / BCODE_BUFFER_BLOCKSIZE == 0 (in Python it would be -1, yes) and (needed_memory - 1) / BCODE_BUFFER_BLOCKSIZE + 1 == 1: 1 block will be allocated. The same thing was done in the old code, a block would be allocated even if the difference was 0.

np-c0mp1ete (approved):

PR #94, opened 2021-06-21
Remove gunpowder icon from BattleInterface

mitrokosta:
Because there is no code for this texture in scripts, there was a white rectangle in the battle interface.


Hammie:
I had a quick look at wether it was used in CoAS. At first I thought is was, but it was actually in changes given to me by @espkk, and not in the vanilla scripts.

So all good for me, and can also confirm it works.

Looking at the changes, I think maybe `m_idCurChargeVBuf` could be made slightly smaller (12 instead of 16 vertices)?

espkk:
> I had a quick look at wether it was used in CoAS. At first I thought is was, but it was actually in changes given to me by @espkk, and not in the vanilla scripts.
> 
> So all good for me, and can also confirm it works.
> 
> Looking at the changes, I think maybe `m_idCurChargeVBuf` could be made slightly smaller (12 instead of 16 vertices)?

Maybe... Tbh, I don't remember this change at all. Most likely someone requested it as I was not interested in new features, just overhauling the engine.
Probably we can move it to COAS compatibility layer?
Btw, I believe we should upload coas scripts to GitHub as well?

mitrokosta:
Thanks, @Hammie ! I forgot to revert the buffer size increase.

espkk (approved):
Well, I'm not sure if we have to completely remove that but let's do this for now to reduce complexity. 
Thanks

PR #100, opened 2021-07-04
Fix a typo in the Steam cmake option test

zeule:


np-c0mp1ete (approved):
Thanks a lot for the fix!

zeule:
> Thanks a lot for the fix!

Thanks! It probably makes sense to use the most comprehend configuration for CI builds? I.e. to include a case with this option turned on :)

np-c0mp1ete:
> Thanks! It probably makes sense to use the most comprehend configuration for CI builds? I.e. to include a case with this option turned on :)

Maybe, but on the other hand it makes sense to build the default configuration which has steam turned off. Also, I guess, running a build with this option turned on wouldn't save us from mistakes like you found.

So, personally I'd probably prefer to keep things as they are for now but if anybody feels differently, feel free to say so.

zeule:
It actually would help in this case, because https://github.com/storm-devs/storm-engine/blob/592a795e781753172a2ff2d5a701fc5db4808427/src/CMakeLists.txt#L5-L7 refers to the proper name adding the preprocessor definition, but without source files and link libraries builds had been failing until this change.

So, I do feel different and suggest to add a third build case to the matrix, msvc-release with the steam option on. The produced artifact would be of use for those who do not want to install Visual Studio but want to try this version, or do not want to compile locally. Also for those like me who can't get anything except a crash on startup from the current versions in Steam.

np-c0mp1ete:
> It actually would help in this case, because
> 
> https://github.com/storm-devs/storm-engine/blob/592a795e781753172a2ff2d5a701fc5db4808427/src/CMakeLists.txt#L5-L7
> 
> refers to the proper name adding the preprocessor definition, but without source files and link libraries builds had been failing until this change.
> So, I do feel different and suggest to add a third build case to the matrix, msvc-release with the steam option on. The produced artifact would be of use for those who do not want to install Visual Studio but want to try this version, or do not want to compile locally. Also for those like me who can't get anything except a crash on startup from the current versions in Steam.

okay, I guess CI build times are not the biggest concern for us and the correctness of different options is more important. Feel free to open a new PR for introducing this build. If somebody will have any issues with it, we'll discuss them there.

zeule:
OK!

PR #101, opened 2021-07-05
Add release build with enabled Steam integration to CI

zeule:


.github/workflows/cibuild.yml
```diff
@@ -21,6 +21,11 @@ jobs:
       fail-fast: false
       matrix:
         configuration: [msvc-debug, msvc-release]#, clang-debug]
+        enable_steam: [false]
```
zeule, line 24: perhaps ON/OFF instead of true/false? ON/OFF is shorter but GitHub actions do not understand them as boolean values.

.github/workflows/cibuild.yml
```diff
@@ -21,6 +21,11 @@ jobs:
...
+        include:
+          - enable_steam: true
+            configuration: msvc-release
+    name: 'build-windows [${{ matrix.configuration}}, steam: ${{ matrix.enable_steam }}]'
```
zeule, line 28: 'build-windows' might be superfluous

zeule (commented):
A couple of strings might be rectified.

espkk:
Is this intended to perform a separate build with steam enabled or to enable it in all builds?
I mean, not all release builds are meant to be steam-integrated
But it's good to have it along with regular one

zeule:
> Is this intended to perform a separate build with steam enabled or to enable it in all builds?
> I mean, not all release builds are meant to be steam-integrated
> But it's good to have it along with regular one

the build matrix is extended with the new "dimension" `stem_enabled`, which is set to false for everything except the `msvc-release` configuration. So it adds a separate (third) build:
https://github.com/storm-devs/storm-engine/pull/101/files#diff-c0686c29bac861bc2f93c992fc7f444c69920f526bc26f49d0dec78384b7ca15R25-R27

The condition needs to be kept in sync with future build matrix changes.

espkk:
Ah, I see. That's good, thanks. I think we can merge it

espkk:
I have no idea why regular build-windows (msvc-debug) and build-windows (msvc-release) are not passing. Any guesses? 

These are push triggers I believe?

zeule:
All three builds passes, but their names changed and settings for branch protection in the repository refers to names that do not exist in this PR, waiting indefinitely for them to complete. 

The default build name for GitHub actions seems to be "<step name> (<comma-separated list of build matrix values>)". When I added "steam_enabled" dimension with values "true" of "false", the automatic build names became "build-windows (msvc-debug, false)", "build-windows (msvc-release, false)", and "build-windows (msvc-release, true)", Those "true" and "false" are meaningless and I defined build name [explicitly](https://github.com/storm-devs/storm-engine/pull/101/files#diff-c0686c29bac861bc2f93c992fc7f444c69920f526bc26f49d0dec78384b7ca15R28). Please see new names in the "Checks" tab or listed below. I also left a note saying my choice of the build names might be not good enough.

Hammie (approved):

espkk:
Thanks! Let's try pushing it anyway

PR #103, opened 2021-07-07
Fix general stability issues

espkk:


Hammie (dismissed):
Looks good to me and should fix a few bugs

PR #105, opened 2021-07-13
Overhauling of build system

espkk:
This is a quite major update ( not in terms of size though :) ) that almost does not affect the code but IMO this is a necessary step for further refactoring and cleanup.
It mostly consists of three parts:
1) implement "autocmake" system called "StormSetup" similar to one that @Hammie added, but considering SE-specific things. It is 'globbing' based but with enabled CONFIGURE_DEPENDS
2) bring back Thunderstom classes registry to get rid of "vmainit". Since we are on static-libs now this is kind of a workaround too. but it is lesser of two evils and also conforms the original architecture
3) move externals to artifactory

It is far from a finished state but I need reviews and comments on that so I can fix and finish the stuff.

Known issues ATM are:
resources are not compiled - to be automated via StormSetup
tests are not compiled - to be automated via StormSetup
clang build is likely not working (I did not test that) - I think StormSetup needs tweaking to support it

Please, take a look and do not hesitate to leave a comment


conanfile.txt
```diff
@@ -1,15 +1,28 @@
...
+
 [imports]
-bin, crashpad_handler.exe -> ./bin
\ No newline at end of file
+bin, crashpad_handler.exe -> ./bin
+bin, crashpad_handler.exe -> ./bin
+bin, 7za.exe -> ./bin
+lib, fmod.dll -> ./bin
+lib, fmodL.dll -> ./bin
+lib, steam_api64.dll -> ./bin
```
WillTer, line 28: How should it works with STORM_ENABLE_STEAM variable disabled by default?

espkk:
I got a comment notification on email but the author already deleted it.
It was a reasonable notice about conanfile.txt imports. Right now I have no idea how to make these imports conditional. So steamapi dll is copied along with both: debug and release versions of fmod dll.
I have several ideas in mind:
1. find a way to do conanfile.txt import depending on cmake configuration
2. use conanfily.py?
3. bring deploying back to cmake
4. trim excess binaries at CI deploy

The problem with debug/release dlls is that we can determine type of build only at generation time for 'cmake' generator; and only at build time for 'cmake-multi' generator. Since I would go with 'cmake-multi' we will have a problem determining correct dependencies in any way.
We can probably do some complex post-build hook (at the buildsystem level), but I feel like CI should handle that much clearer.

Anyway, this is a point for discussion as for now

WillTer:
> I got a comment notification on email but the author already deleted it.
> It was a reasonable notice about conanfile.txt imports. Right now I have no idea how to make these imports conditional. So steamapi dll is copied along with both: debug and release versions of fmod dll.
> I have several ideas in mind:
> 
>     1. find a way to do conanfile.txt import depending on cmake configuration
> 
>     2. use conanfily.py?
> 
>     3. bring deploying back to cmake
> 
>     4. trim excess binaries at CI deploy
> 
> 
> The problem with debug/release dlls is that we can determine type of build only at generation time for 'cmake' generator; and only at build time for 'cmake-multi' generator. Since I would go with 'cmake-multi' we will have a problem determining correct dependencies in any way.
> We can probably do some complex post-build hook (at the buildsystem level), but I feel like CI should handle that much clearer.
> 
> Anyway, this is a point for discussion as for now

I've deleted my comment because it based on misunderstanding (never used the imports section in conanfile and was a bit confused).
About situation in general, in my opinion conanfile.txt is not suited to use in complex situations (and by "complex" i mean anything more than simple dependencies list). But i'm not sure that move to conanfile.py will help with multiconfig problem.

src/apps/config/CMakeLists.txt
```diff
@@ -1,20 +1,4 @@
-project(config)
+# set up this later
+# or even better: completely remove
```
WillTer: Do you have plans to move display preferences to in-game settings?

WillTer:
The second "src" in apps or libs source paths (like `src/apps/ENGINE/src/Core.cpp`) seems overkill, IMHO

src/apps/config/CMakeLists.txt
Hammie: I don't think their are plans for that. The original config tools still work though.

espkk: This extremely old config tool only operates on ini file. It does not depend on anything from the engine sources. At the same time, it requires MFC that no one wants to install. Finally, it does not really require any changes. It can be deployed just like we do it to libs.
What about our plans, I would like to have in-game preferences, yes. But as Hammie noticed we won't have it in the nearest future :)

espkk:
> 
> 
> The second "src" in apps or libs source paths (like `src/apps/ENGINE/src/Core.cpp`) seems overkill, IMHO

I would say the first one is probably excessive :) But the latter has a special meaning: inside the project configured by StormSetup script 'src' denotes that the folder contains the source code, 'include' is getting imported as public headers, 'testsuite' will contain test application for the libraries

src/apps/config/CMakeLists.txt
WillTer: Original config tool mostly useless, because modern resolutions are not supported and you still need to set it up manually in .ini file, where you can configure other options as well. Furthermore, sound volume can be configured in game.

I don't think that this is important for nearest future, but most users (especially those not related to IT) prefer some kind of GUI instead of manual editing config files :)

Edit: I forgot to mention that problem with modern resolutions exists only in old games like CoaS or PotC, so they might need to get updated config tool (or some another way to configure graphics through UI), but that's a very low priority task.

image

espkk: I believe the one from TEHO is slightly modified to support new resolutions. But it is still just a bad way to manage in-game settings.
It is indeed on my to-do list, but we have just a few people working on the engine right now so I would not expect to see it until more important things are done...

WillTer: No problem, just wanted to clarify how it will work in the future (no matter how far away)

Hammie:
I changed the cmake to use `file(RELATIVE_PATH)` instead of a regex because I have the project in my `C++` folder XD

Hopefully I also fixed the build

np-c0mp1ete (dismissed):
Thanks for the improvements!
I don't mind this major rework, but I have a couple of comments:
- git doesn't seem smart enough to detect moving of some files. So the git history is broken for them, and it's impossible to review changes in them (if any) since the whole file is considered new. I think `git mv` might be able to fix that.
- in VS Cmake Targets View paths are displayed like `Public Headers/include` and `Source Files/src`. It would be nice to remove `include` and `src` subfilters.
- could you explain how the VmaInit replacement works now?
- config app and unit tests seem broken at the moment.

cmake/StormSetup.cmake
```diff
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.12.0)
 
 macro(_get_relative_path_folder out path)
-    list(APPEND base_folders ${PUBLIC_INCLUDE_DIRS} ${SRC_DIRS})
+    set(base_folders "${PUBLIC_INCLUDE_DIRS};${SRC_DIRS}")
     list(JOIN base_folders "|" folders_match)
     get_filename_component(dir ${path} DIRECTORY)
-    string(REGEX REPLACE "(.*${CMAKE_CURRENT_SOURCE_DIR}/(${folders_match})\/?)" "" ${out} "${dir}")
+    file(RELATIVE_PATH ${out} "${CMAKE_CURRENT_SOURCE_DIR}" "${dir}")
```
espkk: Breaks VS filters. Is there any reason why we should avoid `REGEX REPLACE`?

espkk:
>     * git doesn't seem smart enough to detect moving of some files. So the git history is broken for them, and it's impossible to review changes in them (if any) since the whole file is considered new. I think `git mv` might be able to fix that.

Sadge. Thanks for noticing, I will try to do it after finishing this PR

>     * in VS Cmake Targets View paths are displayed like `Public Headers/include` and `Source Files/src`. It would be nice to remove `include` and `src` subfilters.

This is brought by a3e7413. @Hammie, thanks for the fixes, but could you please check that?

>     * could you explain how the VmaInit replacement works now?

In thunderstorm static initialization was used to form a list of 'classes'. That was guaranteed by ABI, the same applies to dll. Since we switched to static lib in this fork that rule no longer works so the unreferenced objects are just stripped at the linking stage. Doing /WHOLEARCHIVE is a guarantee to avoid it.
Personally, I don't like relying on such kinds of toolset-specific things, but otherwise, it would be an architecture violation. So it is the lesser of two evils :)

>     * config app and unit tests seem broken at the moment.

I'm about to embed UT in the new setup script. 
What about the config app, I would move it to 'Tools' folder with prebuilt binary and configure deploy to gather it. Let me know if you don't agree.

NB: this is still to be considered as a draft. It is not a 'github draft' because I want to get feedback before it is finished.

cmake/StormSetup.cmake
Hammie: Files paths with `.`, `+`, `(` or `)`. 

The `+` was causing issues for me. So we could still use regexps but then we would need to escape the current source directory first.

I made a mistake with the excluding the `base_folders` though, will try to fix that.

espkk: Ah, I see! I didn't think about it before
But quick googling says we can prepare the path part of that regex with `string(REGEX REPLACE "([][+.*()^])" "\\\\\\1" regex "${string}")`

Hammie: I combined the relative path with the regex search. Should work now, as long as we don't put strange folder names in the project (which we shouldn't).

espkk:
@np-c0mp1ete, I checked the things around git mv. Actually, the files ARE properly moved; git accurately shows that those folders are renamed. 
As far as I understand (from google:)), missing history is a github issue.
Probably this is because the whole folders were renamed... Maybe the reason is just the same that caused Thunderstorm history to be lost?
I can imagine that some kind of force-wildcard-file-rename would fix that, but I have no idea.

espkk:
This is a candidate for merge. The only missing thing is a prebuilt binary for 'config'. Feel free to add it. But I would not consider that as a blocker since I want this PR to be merged ASAP

Hammie (approved):
I think this is ready to be merged. I compiled it and everything seems to still be working :+1:

Probably best to merge these huge changes as soon as possible.

I would like to re-enable the formatting check, but we can do that in a separate PR later.

np-c0mp1ete (approved):
- Regarding the git history, as I mentioned, for some reason some files are marked as new or deleted in git (try `git log -- src/apps/ENGINE/src/token.h`). However, a git client that I use can still somehow show the complete history of the file.
- In the past shared headers and shaders were copied to the build folder, this has been removed in the PR. How is it supposed to work now? 

In general, the build system has definitely got cleaner, and I don't want to prevent you from merging, thanks!

espkk:
>     * Regarding the git history, as I mentioned, for some reason some files are marked as new or deleted in git (try `git log -- src/apps/ENGINE/src/token.h`). However, a git client that I use can still somehow show the complete history of the file.

I believe it shows exactly what is should. And you probably want to use this: 'git log --follow src/apps/ENGINE/src/token.h'
git is really overcomplicated :)

>     * In the past shared headers and shaders were copied to the build folder, this has been removed in the PR. How is it supposed to work now?

Actually I forgot about this, but I believe it's not a crucial step since we do not do any production deploy. Therefore, I will merge it as is for now, but returning to the way how it is supposed to work... right now our bin folder looks like this:
``7za.exe
crashpad_handler.exe
engine.exe
engine.exe.manifest
engine.pdb
fmod.dll
fmodL.dll
steam_api64.dll
test-all.exe
test-all.pdb
util-test.exe
util-test.exe.manifest
util-test.pdb``
you may notice that we do not need util-test for deploy, nor manifest files, steam_api64.dll should appear only for steam vesrion and fmod.dll/fmodL.dll shall be bundled according to the build type.
I think it is a job for CI/CD job to handle this as well as grabbing techniques/shared/whatever.
If we are talking about github CI, we probably need our own actions to handle that and some other things like for example sentry pdb uploading I made for 8cd21191 - it could look better.

In theory, I can take care of this, but I really want to do some additional work on the engine first.
Anyway, we can think about it at any time

PR #108, opened 2021-07-16
Only show supplies and coordinates on world map UI when set

Hammie:


espkk:
Since @mitrokosta mentioned this has an issue for TEHO with empty rect probably compatibility layer can be used to output "N -- --' W -- --'" if coords are not set. 
But considering it's just a string, putting a placeholder into the engine is kind of an unwanted hardcode. 
I would rather modify wdmCoordUpdate to set that coords string from scripts (I guess just removing check `if (isLongitude || isLatitude)` will do the job). It won't break compatibility with the original TEHO engine at this point.
@mitrokosta, can you investigate this, please

mitrokosta:
@espkk Your solution partially fixes it. The coordinates appear after a 1-second delay.

espkk:
> @espkk Your solution partially fixes it. The coordinates appear after a 1-second delay.

Does not that mean that the event used for that indication does not really fit for that purpose? I mean, the delay means the correct coords are also set after that delay, right?
Maybe we have some more appropriate event to set them asap when entering worldmap?

Hammie:
I've re-added the placeholder coordinates for TEHO, so for now things should work like they did previously

mitrokosta (dismissed):
It's fine now!

espkk (dismissed):

espkk (approved):

PR #109, opened 2021-07-18
[CoaS] Fix crash when opening Index of Terms in QuestBook on ru-locale

WillTer:
Backward compatibility for CoaS. Some texts in russian do not fit to the buffer that results in breaking multibyte utf symbol and crash when trying to read that symbol.

A better way to fix this is to rewrite questbook functions (and all XInterface lib) to use more high-level approach to work with strings (without raw buffers and etc.). But that needs much more time, so i decided to provide quickfix just for now.


np-c0mp1ete (approved):
Thanks for the fix!

espkk (approved):
Thank you

PR #111, opened 2021-07-19
Refactor QuestFileReader class

WillTer:
Changes no one asks for :)

I've made a massive refactoring of QuestFileReader class. The main goal is to simplify parser and update all code to actual codestyle and modern features.
I mostly use std::string_view as text container to prevent unnecessary copies.
Buffers was changed frow raw C-style arrays to std::vector<char>. The main reason of that choice is to make buffer size dynamic and prevent errors that can occur with hardcoded size (when it's insufficient).

This request is some kind of draft. Planned changes are done and the game works fine at first glance (tested on russian version of CoaS and TEHO) but I want to do more tests (for example, run and play PotC).

I would be glad to see comments and remarks on the chosen approach. This is especially true for comments - i'm not native english speaker, and want them to be clear and understandable.

### DIFF
--- a/src/libs/Xinterface/Nodes/xi_questtitles.cpp
+++ b/src/libs/Xinterface/Nodes/xi_questtitles.cpp
@@ -332,7 +332,7 @@ void CXI_QUESTTITLE::SetNewTopQuest(ATTRIBUTES *pA, int topNum)
     }
     m_nCommonQuantity = aq;
 
-    char param[256];
+    std::string param;
     // if(ptrOwner->QuestFileReader())
     {
         // ptrOwner->QuestFileReader()->InitQuestsQuery();
@@ -369,22 +369,23 @@ void CXI_QUESTTITLE::SetNewTopQuest(ATTRIBUTES *pA, int topNum)
             const char *pTmpQuestRecordID = pAttr->GetAttribute("LogName");
             if (!pTmpQuestRecordID)
                 pTmpQuestRecordID = pAttr->GetThisName();
-            if (ptrOwner->QuestFileReader()->GetQuestTitle(pTmpQuestRecordID, pAttr->GetThisName(), sizeof(param) - 1,
-                                                           param))
+            if (ptrOwner->QuestFileReader()->GetQuestTitle(pTmpQuestRecordID, pAttr->GetThisName(), param))
             {
-                const int titleSize = strlen(param);
+                const size_t titleSize = param.size();
                 if (titleSize == 0)
                     m_strList[i].lineQuantity = 0;
                 else
                 {
-                    char lineName[sizeof(param)];
-                    auto *pstr = param;
+                    std::string lineName(titleSize, '\0');
+                    auto *pstr = param.data();
                     int ln = 0;
-                    while (GetLineNext(m_idFont, pstr, lineName, sizeof(lineName)))
+                    /// @todo There is no need to get size here, instead there
+                    /// will be better to use std::vector in GetLineNext
+                    while (GetLineNext(m_idFont, pstr, lineName.data(), lineName.size()))
                     {
-                        const auto len = strlen(lineName) + 1;
+                        const auto len = lineName.size() + 1;
                         m_strList[i].name[ln] = new char[len];
-                        memcpy(m_strList[i].name[ln], lineName, len);
+                        memcpy(m_strList[i].name[ln], lineName.data(), len);
                         ln++;
                     }
                     m_strList[i].lineQuantity = ln;
--- a/src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
+++ b/src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
@@ -9,9 +9,11 @@
 
 #include "storm_assert.h"
 
-namespace {
+namespace
+{
 
-enum class Token {
+enum class Token
+{
     Invalid,
     Quest,
     Text,
@@ -19,18 +21,15 @@ enum class Token {
     Unknown,
 };
 
-const std::map<std::string_view, Token> kTokenMap = {
-    {"#QUEST", Token::Quest},
-    {"#TEXT", Token::Text}
-};
+const std::map<std::string_view, Token> kTokenMap = {{"#QUEST", Token::Quest}, {"#TEXT", Token::Text}};
 
 /// @brief Read text until whitespace or newline
 /// @param[in] str Text buffer to read from
 /// @param[in,out] it An iterator pointing to place where reading needs to be started,
-///                   will be set to character right after readed text in str
-/// @return Readed text
+///                   will be set to the character right after the text that was read from str
+/// @return Text that was read
 /// @note If there is whitespaces in the beginning, they will be trimmed before reading
-std::string_view ReadWord(const std::string_view & str, std::string_view::const_iterator &it)
+std::string_view ReadWord(const std::string_view &str, std::string_view::const_iterator &it)
 {
     auto begin = std::find_if(it, str.cend(), [](const char ch) { return !isspace(static_cast<unsigned char>(ch)); });
     auto end = std::find_if(begin, str.cend(), [](const char ch) { return isspace(static_cast<unsigned char>(ch)); });
@@ -43,9 +42,10 @@ std::string_view ReadWord(const std::string_view & str, std::string_view::const_
 /// @brief Read text until new token starts ('#' symbol)
 /// @param[in] str Text buffer to read from
 /// @param[in,out] it An iterator pointing to place where reading needs to be started,
-///                   will be set to '#' character right after readed text in str
-/// @return Readed text
-/// @note If there is whitespaces in the beginning, they will be trimmed before reading. Trailing whitespaces will be trimmed too.
+///                   will be set to the '#' character right after the text that was read from str
+/// @return Text that was read
+/// @note If there is whitespaces in the beginning, they will be trimmed before reading. Trailing whitespaces will be
+/// trimmed too.
 std::string_view ReadText(const std::string_view &str, std::string_view::const_iterator &it)
 {
     auto begin = std::find_if(it, str.cend(), [](const char ch) { return !isspace(static_cast<unsigned char>(ch)); });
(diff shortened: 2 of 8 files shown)


src/libs/Xinterface/Stringservice/strservice.cpp
```diff
@@ -1425,24 +1425,24 @@ uint32_t _DialogAssembleStr(VS_STACK *pS)
         return IFUNCRESULT_FAILED;
     const char *pcID = pDat->GetString();
 
-    char param[4096];
+    std::vector<char> param;
```
np-c0mp1ete: why not std::string?

src/libs/Xinterface/QuestFileReader/QuestFileReader.h
```diff
@@ -1,14 +1,22 @@
...
 
-class QUEST_FILE_READER
+namespace storm
+{
+class QuestFileReader
 {
   public:
     struct UserData
     {
+        /// @note Constructor needs for emplace_back
```
np-c0mp1ete: `is needed` instead of `needs`

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
         return;
     }
-    // create a buffer for it
-    char *pBuf = new char[filesize + 1];
-    Assert(pBuf);
-    // read into this buffer from a file
-    if (!fio->_ReadFile(fileS, pBuf, filesize))
+
+    /// Create a buffer
+    std::vector<char> buffer(filesize + 1);
```
np-c0mp1ete: why not std::string?

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+
+        if (patternStart == it) ///< Found an incorrect pattern or EOF
+            continue;
+
+        std::string insertStr = GetInsertStringByID(insertID, userData);
+        buffer.insert(buffer.cend(), insertStr.cbegin(), insertStr.cend());
     }
-    pBuf[nDst] = 0;
+
+    buffer.push_back(0); ///< Workaround for methods that works with raw char* buffer
```
np-c0mp1ete: `that work`

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
-                nIns = 0;
-            }
-            else
-            {
-                pBuf[nDst] = pSrc[nSrc];
-                nDst++;
-            }
-        }
+        auto [insertID, patternStart] = ReadUserDataID(src, it);
+        buffer.insert(buffer.cend(), textStart, patternStart); ///< Insert all text before pattern to buffer
```
np-c0mp1ete: `into a buffer`

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+    const auto tokenString = ReadWord(str, it);
+    const auto token = kTokenMap.contains(tokenString) ? kTokenMap.at(tokenString) : Token::Unknown;
+
+    const auto id = ReadWord(str, it);
+    const auto text = ReadText(str, it);
+
+    return {token, id, text};
+};
+
+/// @brief Read UserDataID, that placed in special pattern
```
np-c0mp1ete: `is placed`

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+/// @param[in] str Text buffer to read from
+/// @param[in,out] it An iterator pointing to place where reading needs to be started,
+///                   will be set to character right after readed text in str
```
np-c0mp1ete: maybe `will be set to the character right after the end of str`?

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
-    }
-    pOutBuf[i] = 0;
+    return std::string_view(text.begin(), end); ///< We must use text.begin() because end is now belongs to text
 }
 
-static const char *GetNextString(const char *&pInStr)
+/// @brief Read token with all related data
+/// @param[in] str Text buffer to read from
+/// @param[in,out] it An iterator pointing to place where reading needs to be started,
+///                   will be set to character right after readed text in str
```
np-c0mp1ete: maybe `will be set to the character right after the end of str`?

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
-    return static_cast<long>(ch - '0');
+    return std::string_view(begin, end);
 }
 
-// Functions for working with strings
-static long GetLongFromString(char *&pInStr)
+/// @brief Read text until new token starts ('#' symbol)
+/// @param[in] str Text buffer to read from
+/// @param[in,out] it An iterator pointing to place where reading needs to be started,
+///                   will be set to '#' character right after readed text in str
```
np-c0mp1ete: maybe `will be set to the '#' character right after the end of str`?

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+/// @return Readed text
```
np-c0mp1ete: `Text that was read`

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+
+    Unknown,
+};
+
+const std::map<std::string_view, Token> kTokenMap = {{"#QUEST", Token::Quest}, {"#TEXT", Token::Text}};
+
+/// @brief Read text until whitespace or newline
+/// @param[in] str Text buffer to read from
+/// @param[in,out] it An iterator pointing to place where reading needs to be started,
+///                   will be set to character right after readed text in str
```
np-c0mp1ete: maybe `will be set to the character right after the end of str`?

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
+/// @return Readed text
```
np-c0mp1ete: `text read`

src/libs/Xinterface/Nodes/xi_questtitles.cpp
```diff
@@ -369,22 +369,21 @@ void CXI_QUESTTITLE::SetNewTopQuest(ATTRIBUTES *pA, int topNum)
...
             {
-                const int titleSize = strlen(param);
+                const size_t titleSize = param.size();
                 if (titleSize == 0)
                     m_strList[i].lineQuantity = 0;
                 else
                 {
-                    char lineName[sizeof(param)];
-                    auto *pstr = param;
+                    std::vector<char> lineName(titleSize);
```
np-c0mp1ete: std::string?

src/libs/Xinterface/Nodes/xi_questtitles.cpp
```diff
@@ -332,7 +332,7 @@ void CXI_QUESTTITLE::SetNewTopQuest(ATTRIBUTES *pA, int topNum)
     }
     m_nCommonQuantity = aq;
 
-    char param[256];
+    std::vector<char> param;
```
np-c0mp1ete: std::string?

WillTer:
I was too optimistic about testing on PotC :D
With some script tweaking I can start the game and even go to the Oxbay City. So I've tested my changes on tutorial quest (with some tweaks here too, because PotC scripts use quite different approach to show quest texts), and all works fine.
In addition to that I've looked at quest list variable in debug, and as I see, all quest texts loaded correctly.

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
```diff
@@ -1,644 +1,392 @@
...
-    }
-
-    if (m_pFileBuf == nullptr || m_pFileBuf[0] == 0)
-    {
-        return false;
-    }
-    return true;
+namespace storm
+{
+QuestFileReader::QuestFileReader()
```
Hammie: `QuestFileReader` constructor and destructor are no longer needed. So you can remove them if you want.

Hammie (dismissed):
Looks good to me. 

Nobody may have asked for these changes, but we welcome the improvements either way :+1:

src/libs/Xinterface/Nodes/xi_questtitles.cpp
```diff
@@ -369,22 +369,21 @@ void CXI_QUESTTITLE::SetNewTopQuest(ATTRIBUTES *pA, int topNum)
...
                     m_strList[i].lineQuantity = 0;
                 else
                 {
-                    char lineName[sizeof(param)];
-                    auto *pstr = param;
+                    std::vector<char> lineName(titleSize);
+                    auto *pstr = param.data();
                     int ln = 0;
-                    while (GetLineNext(m_idFont, pstr, lineName, sizeof(lineName)))
+                    while (GetLineNext(m_idFont, pstr, lineName.data(), lineName.size())) /// @todo
```
np-c0mp1ete: is to-do still valid? maybe add a comment what still needs to be done then?

np-c0mp1ete (commented):
Thanks a lot for the rework, updating the code to the modern C++ is always appreciated.
I left mostly grammar related comments since you explicitly asked for it.
If you feel std::vector<char> is more suitable here than std::string, I'm fine with that as well.

src/libs/Xinterface/Stringservice/strservice.cpp
WillTer: Because std::string contains const char*, that is not fully suitable for buffer (because of const nature)

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
WillTer: Not really 'the end of str', because str is text that we reading from and we not necessarily reached the end of it

src/libs/Xinterface/Nodes/xi_questtitles.cpp
WillTer: Yeah, forgot to complete a comment

src/libs/Xinterface/Stringservice/strservice.cpp
WillTer: The only problem with a vector is that it doesn't automaticaly add null terminator to the end, but that is not different from using raw char arrays)

WillTer: Oh, I forgot that basic_string have data() method too, that can return non-const array, so the comments I write before is no longer relevant xD
Maybe it will be better to replace vector by string

src/libs/Xinterface/QuestFileReader/QuestFileReader.cpp
np-c0mp1ete: ah, I see. maybe then `will be set to the character right after the text that was read from str`?

WillTer: I fixed that and make fixup of last commits

np-c0mp1ete (approved):

PR #114, opened 2021-07-20
blender-gm-import: change weight add type

Artess999:
Seems to work better on some models, for example, monkey


np-c0mp1ete (approved):
Thanks!

PR #116, opened 2021-07-20
Fix screen size for PotC compatibility

Hammie:
_Pirates of the Caribbean_ assumed a different, smaller screen size (for configs, not actual screen size)


espkk (approved):
Looks good, thanks!

PR #128, opened 2021-07-25
Fix crash when text attribute is not set

WillTer:
Needed mainly for backward compatibility

sText is std::string and trying to construct it with nullptr is UB.


espkk (approved):
looks good, thanks!

PR #129, opened 2021-07-25
Fix UserData obtaining for quest headers

WillTer:
Sorry, I screwed up when I did the refactoring :(


src/libs/xinterface/src/QuestFileReader/QuestFileReader.cpp
```diff
@@ -167,7 +167,7 @@ std::string QuestFileReader::GetRecordText(const std::string_view &questID, cons
     const auto &unwrappedText = quest.texts[textNum.value()];
     AssembleStringToBuffer(unwrappedText.str.c_str(), buffer, questData_);
 
-    return std::string(buffer.data(), buffer.size());
+    return buffer;
```
WillTer, line 170: This is just an accompanying fix. I've changed vector to string, but forgot about that line. No errors here, just useless conversion.

np-c0mp1ete (approved):
thanks for the fix!

espkk (approved):

PR #133, opened 2021-07-26
Simplify ScrollImage code a little

Hammie:
Code is still largely a mess, but there should be a few less raw pointer arrays now.


espkk (approved):
looks good, thanks!

that non-const stuff for everything including literals is really terrible, just like the rest of xinterface :)

PR #135, opened 2021-07-27
Fix crash when string attribute in BI is not exists

WillTer:
Comparing std::string with nullptr results to crash


Hammie (approved):
Since this code is used to detect changes in attributes, I agree not having the attribute should result in "not changed".

Thanks!

espkk (approved):
Looks good, thanks!

PR #138, opened 2021-07-28
fix handling _GetExecutableDirectory return value

espkk:
_GetExecutableDirectory uses SDL_GetBasePath that returns utf-8 path. 
That is not preserved in two places leading to crashes if the engine or documents folders contains non-locale symbols in their path


np-c0mp1ete (dismissed):
Looks fine, thanks!
@Hammie feel free to merge the PR, if you don't have any comments.

espkk:
whoops. I accidentaly grabbed a change from another branch. should be cleaner now

Hammie (approved):

PR #141, opened 2021-07-29
Maintenance fixes

espkk:
A few general bugfixes + a lot of SCA fixes


src/libs/geometry/src/geometry.cpp
```diff
@@ -229,7 +229,7 @@ void *GEOM_SERVICE_R::malloc(long bytes)
 
 void GEOM_SERVICE_R::free(void *ptr)
 {
-    delete ptr;
+    delete[] static_cast<char*>(ptr);
```
WillTer, line 232: What purpose of casting before deletion?

src/libs/island/src/ISLAND.cpp
```diff
@@ -775,7 +775,7 @@ bool ISLAND::Mount(const std::string_view &fname, const std::string_view &fdir,
     if (pNode)
         seabed_id = pNode->Unlink2Model();
     else
-        core.Trace("Island: island %s has no sea bed, check me!", fname);
+        core.Trace("Island: island %s has no sea bed, check me!", std::string(fname).c_str());
```
WillTer, line 778: Why not `fname.data()`?

src/libs/xinterface/src/QuestFileReader/QuestFileReader.cpp
```diff
@@ -265,15 +265,15 @@ void QuestFileReader::SetQuestTextFileName(const std::string_view &fileName)
     auto fileS = fio->_CreateFile(fileName.data(), std::ios::binary | std::ios::in);
     if (!fileS.is_open())
     {
-        core.Trace("WARNING! Can`t open quest log file %s", fileName);
+        core.Trace("WARNING! Can`t open quest log file %s", std::string(fileName).c_str());
```
WillTer, line 268: `fileName.data()`?

src/libs/geometry/src/geometry.cpp
espkk: this is just not allowed by standard: https://eel.is/c++draft/expr.delete#footnote-69
practically, usually there is no difference, nevertheless it is still UB and treated as a misusage by most modern compilers and SCA tools

src/libs/island/src/ISLAND.cpp
espkk: because std::string_view has no guarantee that the data is null-terminated. https://eel.is/c++draft/string.view#access-14
actually, that's the key feature of string_view

src/libs/xinterface/src/QuestFileReader/QuestFileReader.cpp
espkk: see above

src/libs/island/src/ISLAND.cpp
espkk: to the point, ideally we should switch to spdlog/fmt formatting completely; that would also solve such kind of issues. recent spdlog update targets fmt 8 support that I believe can help a lot with logging refactoring since it brings stable consteval formatting lexic parser

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -5206,7 +5206,7 @@ bool COMPILER::BC_Execute(uint32_t function_code, DATA *&pVReturnResult, const c
                     return false;
                 if (pVDst->GetType() != VAR_REFERENCE)
                 {
-                    SetError("'%s' isnt reference", real_var->name.c_str());
+                    SetError("Local variable is not reference");
```
np-c0mp1ete, line 5203: why did you remove the var name?

src/apps/ENGINE/src/s_dbg_watcherlist.cpp
```diff
@@ -18,25 +16,22 @@ WATCHER_LIST::WATCHER_LIST(HWND hwnd, HINSTANCE hinst)
...
+        long n = 0;
         sprintf_s(buffer, "E%d", n);
         if (ini->ReadString(nullptr, buffer, buffer, sizeof(buffer), ""))
         {
-            if (n < 11) //~!~
-            {
-                SetItemText(n, 0, buffer);
-                SetItemText(n, 1, CDebug.ProcessExpression(buffer));
-            }
+            SetItemText(n, 0, buffer);
```
np-c0mp1ete, line 33: what was the reason for the n < 11 check?

src/libs/lighter/src/LGeometry.h
```diff
@@ -18,12 +18,14 @@ class LGeometry
...
         Object()
+            : model(0)
         {
             name = nullptr;
             nameReal = nullptr;
             m = nullptr;
             lBufSize = 0;
-        };
+        }
+        ;
```
np-c0mp1ete, line 28: is it on purpose?

Hammie (dismissed):
A bunch of TODO's were added, so it seems we have more to do ;)

I do like seeing a lot less uninitialized variables though :+1:

I also tested it with CoAS and everything seemed to be working.

Hammie:
@np-c0mp1ete Do you have anything to add, or can we go ahead and merge it?

src/apps/ENGINE/src/s_stack.cpp
```diff
@@ -29,10 +29,6 @@ void S_STACK::Release()
 
 DATA *S_STACK::Push(DATA *pdataclass)
 {
-    if (Data_num > 1000)
```
np-c0mp1ete, line 32: what was the purpose of this check?

src/libs/sailors/src/Sailors.cpp
```diff
@@ -34,7 +35,8 @@ ShipMan::ShipMan()
     jumpSpeedX = jumpSpeedY = 0;
 
     // UN//GUARD_SAILORS
-};
+}
+;
```
np-c0mp1ete, line 39: maybe just remove the semicolon?

src/libs/sailors/src/Sailors.cpp
```diff
@@ -1025,7 +1028,8 @@ Sailors::Sailors()
     IsOnDeck = 0;
 
     // UN//GUARD_SAILORS
-};
+}
+;
```
np-c0mp1ete, line 1032: remove?

src/libs/sailors/src/Sailors.h
```diff
@@ -50,10 +50,12 @@ struct ShipState
     SEA_BASE *sea;
 
     ShipState()
+        : sea(nullptr)
     {
         mode = SHIP_SAIL;
         dead = false;
-    };
+    }
+    ;
```
np-c0mp1ete, line 58: remove?

src/libs/sea_ai/src/AIShip.h
```diff
@@ -177,12 +177,14 @@ class AIShip : public VAI_INNEROBJ
...
-        AI_POINT(){};
+        AI_POINT() = default;
 
         AI_POINT(VAI_INNEROBJ *_pObj)
+            : fAngle(0)
         {
             pObj = _pObj;
-        };
+        }
+        ;
```
np-c0mp1ete, line 187: remove?

src/libs/sea_cameras/src/Common_Camera.h
```diff
@@ -102,11 +102,13 @@ class COMMON_CAMERA : public Entity
...
 
     COMMON_CAMERA()
+        : eidObject(0), pAIObj(nullptr), pACharacter(nullptr)
     {
         bOn = false;
         bActive = false;
         fPerspective = 1.285f;
-    };
+    }
+    ;
```
np-c0mp1ete, line 111: remove?

np-c0mp1ete (commented):
Thanks for the PR! I added some minor comments/questions, but in general the code looks fine, feel free to merge. I'm not necessarily insisting on addressing my comments.

src/apps/ENGINE/src/compiler.cpp
espkk: We cannot get its name from here... calling this code will lead to a crash

src/apps/ENGINE/src/s_dbg_watcherlist.cpp
espkk: it's a copy-paste from the loop below. it makes no sense here and looks weird here

src/libs/lighter/src/LGeometry.h
espkk: no. this is auto-formatting (maybe from R#). the semicolon is excessive so I believe the formatter inserted a newline after a finished token

src/apps/ENGINE/src/s_stack.cpp
espkk: this is typical for engine developers to check for abnormal behavior. it's likely been used during development to detect recursiveness, etc.

espkk:
These newlines before surplus semicolons are the result of R# formatting around processed core areas... I'll check if there is a way to auto-remove all of them

espkk:
I'll remove redundant semicolons with the next global reformatting

PR #146, opened 2021-08-01
Lost device initial support

espkk:
Fixes alt+tab/UAC/ctrl+alt+tab related crashes. Recently(sdl2), window mode got also partially affected.
This does not restore everything correctly so far. But user can do save;load to recover instead of getting crash.
More fixes (basically restoring render buffers for various entities) will be done with time.
But I think it makes sense to merge it now to have an ability to recover without losing progress.

### DIFF
--- a/src/libs/renderer/src/sdevice.cpp
+++ b/src/libs/renderer/src/sdevice.cpp
@@ -27,6 +27,31 @@ CREATE_SCRIPTLIBRIARY(DX9RENDER_SCRIPT_LIBRIARY)
 
 DX9RENDER *DX9RENDER::pRS = nullptr;
 
+class LostDeviceSentinel : public SERVICE
+{
+    void RunStart() override
+    {
+        if (auto d3d9 = static_cast<IDirect3DDevice9 *>(DX9RENDER::pRS->GetD3DDevice()))
+        {
+            switch (d3d9->TestCooperativeLevel())
+            {
+            case D3DERR_DEVICENOTRESET:
+                if (!DX9RENDER::pRS->ResetDevice())
+                {
+                    core.stopFrameProcessing();
+                }
+                break;
+
+            case D3DERR_DEVICELOST:
+                core.stopFrameProcessing();
+                break;
+            }
+        }
+
+    }
+};
+CREATE_SERVICE(LostDeviceSentinel)
+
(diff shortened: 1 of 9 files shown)


src/libs/renderer/src/sdevice.cpp
```diff
@@ -3441,7 +3466,18 @@ HRESULT DX9RENDER::DrawPrimitive(D3DPRIMITIVETYPE dwPrimitiveType, UINT StartVer
 HRESULT DX9RENDER::Release(IUnknown *pObject)
 {
     if (pObject)
-        return pObject->Release();
+    {
+        if (*(void **)pObject == nullptr)
```
np-c0mp1ete, line 3470: why do we need this code?

src/libs/weather/src/SKY.cpp
```diff
@@ -65,207 +88,175 @@ void SKY::SetDevice()
...
+        iSkyIndexID = pRS->CreateIndexBuffer(20 * 3 * 2);
+    }
 
     const auto fpdelta = 1.0f / 1024.0f;
     const auto fpdx = 1.0f - fpdelta; //.25f - fpdelta;
     const auto fp0 = fpdelta;
-    auto fp1 = 0.25f + fpdelta;
-    auto fp2 = 0.5f + fpdelta;
-    auto fp3 = 0.75f + fpdelta;
+    //auto fp1 = 0.25f + fpdelta;
```
np-c0mp1ete, line 184: is it on purpose?

np-c0mp1ete (commented):
Thanks for the PR!

src/libs/weather/src/SKY.cpp
espkk: Yeah since those are not used. I think it's good practice for this project to comment out dead code instead of removing it when you have no idea why it is not used actually 😆 I'd remove this after revising all that math (one day 🤔 )

src/libs/renderer/src/sdevice.cpp
espkk: This is a kind of debug code that does not harm but should be removed in the future. 
Basically, this check allowed me to detect already released COM objects. And since I'm not done with the feature yet I think it's better to leave it for now.
Of course, if it's not causing issues for anyone...

np-c0mp1ete (approved):

PR #149, opened 2021-08-03
Prevent crash when attemting to load save with utf-8 path

espkk:
Workaround; should be revised later


Hammie (approved):

Hammie:
I made an issue to keep track of the `TODO`: #150

PR #153, opened 2021-08-09
Use IDs instead of pointers for savegame textures

Hammie:


espkk:
If I understand this change correctly that is what I have planned to do for a long time!
We should also remove 'ptr' type from the compiler as this was its only purpose. And I think the corresponding changes should be done to sd-teho's and sd-teho-public's develop branch to conform with the changes

Hammie:
It was causing issues with the way PotC handled its screenshots for save games, and I noticed it was one of the few things still dealing with texture pointers.

I'd have to check if `ptr` is used anywhere else, but it would indeed be nice if we could remove it.

espkk:
> It was causing issues with the way PotC handled its screenshots for save games, and I noticed it was one of the few things still dealing with texture pointers.
> 
> I'd have to check if `ptr` is used anywhere else, but it would indeed be nice if we could remove it.

I believe it is used only for save_load interface. PTR type is something I was quickly added as a workaround to run on x64. Of course, introducing handles would be a better solution. That's what I planned to do eventually

np-c0mp1ete:
I merged the update for the TEHO scripts: https://github.com/storm-devs/sd-teho-public/pull/6

espkk (approved):

PR #155, opened 2021-08-13
Blender gm export

Artess999:
This is first version, early draft
Only one mesh is exported

Suitable for ammo and characters


espkk (approved):
Thanks!

PR #160, opened 2021-08-18
bi: fix double destruction

espkk:


src/libs/battle_interface/src/sea/ibattle.cpp
```diff
@@ -75,9 +75,6 @@ BATTLE_INTERFACE::~BATTLE_INTERFACE()
 {
     // STORM_DELETE(m_pMessageIcons);
     STORM_DELETE(m_pShipIcon);
-    m_TextArray.clear();
```
np-c0mp1ete, line 78: what is the other place where they get destructed?

espkk: In their destructors :)

espkk: `BILinesInfo::~BILinesInfo()
{
    Release();
}`
`BIImagesInfo::~BIImagesInfo()
{
    Release();
}`
The erased code is in `BATTLE_INTERFACE` class that contains both, so it will release resources again. Since `m_ImagesInfo` not nulling pointers after delete, so it will lead to a crash.
Btw, I have no idea how this worked before. This was probably unused in COAS/TEHO, but I hadn't check specifically.

np-c0mp1ete: ah, makes sense.

np-c0mp1ete (approved):
Thanks, all good on my side!

PR #161, opened 2021-08-18
Add fallback support for raw texture formats (e.g. TGA)

espkk:


src/libs/renderer/src/sdevice.cpp
```diff
@@ -1669,6 +1676,30 @@ bool DX9RENDER::TextureLoad(long t)
     return true;
 }
 
+bool DX9RENDER::TextureLoadUsingD3DX(const char* path, long t)
+{
+    // TODO: reimplement the whole thing in a tidy way
+    IDirect3DTexture9 *pTex;
+    if(CHECKD3DERR(D3DXCreateTextureFromFileA(d3d9, path, &pTex)))
+    {
+        delete Textures[t].name;
```
np-c0mp1ete, line 1685: shouldn't it be delete[]?

src/libs/renderer/src/sdevice.cpp
```diff
@@ -1669,6 +1676,30 @@ bool DX9RENDER::TextureLoad(long t)
     return true;
 }
 
+bool DX9RENDER::TextureLoadUsingD3DX(const char* path, long t)
+{
+    // TODO: reimplement the whole thing in a tidy way
+    IDirect3DTexture9 *pTex;
+    if(CHECKD3DERR(D3DXCreateTextureFromFileA(d3d9, path, &pTex)))
```
np-c0mp1ete, line 1683: shouldn't we call the Unicode function and convert our utf8 path to utf16?

src/libs/renderer/src/sdevice.cpp
espkk: It should, but it's a subject for a different PR. I mentioned that in the code. This is copied from the original code and the original code is a mess.
So, let's separate a feature from existing issues. I will take care of this when I have time...

src/libs/renderer/src/sdevice.cpp
espkk: The same as above... yes, we should, as well as not capping length with MAX_PATH aka 260(used in the original code). IIRC Unicode length on windows can be up to (2^16-1)/2
This is to be resolved with a separate maintaining PR

np-c0mp1ete: but I guess if we leave it as it is, textures won't be loaded if there are non-ascii characters in the texture path, right?

espkk: In this case, no, the path is relative to %CD%. And since there is no Unicode support for most asset paths I highly doubt it may cause any issues with the existing games :)

espkk: Most issues start when using %USERPROFILE%. I.e. for logging, savedata, sentry-db, etc.
Everything that lays in the %CD% subfolders normally should not use absolute paths at all. And therefore should not cause issues with Unicode path. At least until we want to introduce it in the subfolder structure... do we?

np-c0mp1ete: personally, I'd prefer to use absolute paths for everything.
okay, in this case it's fine to leave it as it is for now. thanks for the clarification!

np-c0mp1ete (approved):

PR #164, opened 2021-08-19
Fix/xinterface issues

espkk:


np-c0mp1ete (approved):
Haven't looked deeply in the code logic, but the code itself looks fine.
I wonder if there might be any issues with TEHO since @ugeen4 had left a comment regarding his fix that you removed.

espkk:
> 
> 
> Haven't looked deeply in the code logic, but the code itself looks fine.
> I wonder if there might be any issues with TEHO since @ugeen4 had left a comment regarding his fix that you removed.

I indeed removed that check. it seems a bit illogic to me. anyway, it's good you mentioned him so we can return to it later

PR #176, opened 2021-08-22
Remove character in location limit

mitrokosta:
By changing static arrays to vectors I removed the engine's limit of 64 characters per location. Scripts have their own limit in form of ``#define MAX_CHARS_IN_LOC	64``, but this can be changed to any value, 128 for example, or the scripts could be rewritten to remove any limits whatsoever.

### DIFF
--- a/src/libs/location/src/Player.cpp
+++ b/src/libs/location/src/Player.cpp
@@ -632,9 +632,8 @@ Player *Player::FindAttackCharacter()
 {
     auto *const location = GetLocation();
     // Find the surrounding characters
-    static Supervisor::FindCharacter fndCharacter[MAX_CHARACTERS];
-    static long num = 0;
-    if (!location->supervisor.FindCharacters(fndCharacter, num, this, CHARACTER_ATTACK_DIST * 1.1f))
+    auto fndCharacter = location->supervisor.FindCharacters(this, CHARACTER_ATTACK_DIST * 1.1f);
+    if (fndCharacter.empty())
         return nullptr;
     // Choosing the best
     float minDst;
@@ -645,7 +644,7 @@ Player *Player::FindAttackCharacter()
     const auto cdx = sinf(ay);
     const auto cdz = cosf(ay);
     long j = -1;
-    for (long i = 0; i < num; i++)
+    for (size_t i = 0; i < fndCharacter.size(); i++)
     {
         // Character
         auto &fc = fndCharacter[i];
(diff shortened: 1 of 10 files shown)


np-c0mp1ete (approved):
Looks good to me, thanks for the PR!

espkk:
Hey, thanks for the PR.
Have you tested this with, for instance, 128 characters?

src/libs/location/src/Player.cpp
```diff
@@ -771,8 +770,9 @@ void Player::FireFromShootgun()
                 auto *const e = EntityManager::GetEntityPointer(collide->GetObjectID());
                 if (e && e != this)
                 {
-                    long n, nm;
-                    for (n = 0, nm = location->supervisor.numCharacters; n < nm; n++)
+                    long nm;
+                    size_t n;
+                    for (n = 0, nm = location->supervisor.character.size(); n < nm; n++)
```
espkk, line 775: since you modifying the code anyway, nm is immutable, consider replacing it with `location->supervisor.character.size()` here and below:
`for (n = 0; n < location->supervisor.character.size(); n++)`
`if (n == location->supervisor.character.size())`

src/libs/location/src/Supervisor.cpp
```diff
@@ -42,47 +41,47 @@ Supervisor::~Supervisor()
...
     {
         character[i].c->Move(dltTime);
         character[i].c->colMove = 0.0f;
         character[i].c->isCollision = false;
     }
     // calculate the distances, and determine the interacting characters
     long chr = 0;
-    long i;
-    for (i = 0; i < numCharacters - 1; i++)
+    size_t i;
```
espkk, line 75: `i` value is not used outside the loop, it should be declared in the loop

espkk (approved):
Overall looks good to me👍 but I want to know if the supervisor part is tested with all that

mitrokosta:
> Have you tested this with, for instance, 128 characters?

I did, it works, but FPS drops to 40 on my laptop, after 128 characters are logined.

image


I'll fix other issues, give me a moment

espkk:
> 
> 
> > Have you tested this with, for instance, 128 characters?
> 
> I did, it works, but FPS drops to 40 on my laptop, after 128 characters are logined.
> 
> image
> 
> I'll fix other issues, give me a moment

Good!
Of course, it drops, characters animation is calculated on a single CPU core :)

PR #179, opened 2021-08-26
Remove unused, redundant and write-only classes

mitrokosta:


np-c0mp1ete (dismissed):
Removing dead code is always great, thanks!

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -1347,27 +1339,22 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
...
-            SLIBHOLDER *pH;
-            bool bFound;
             VMA *pClass;
 
             //-----------------------------------------------------
             // check if already loaded
-            bFound = false;
-            for (n = 0; n < LibriaryFuncs.GetClassesNum(); n++)
+            auto name = Token.GetData();
+            auto comparator = [name](const auto &library)
```
espkk, line 1349: please, use storm::iStrComparator instead

espkk (changes requested):

espkk:
Please, check the review point above. We want to achieve case-insensitive comparators to refer to the same class.
Also, try to avoid using non-hierarchical branch names

src/apps/ENGINE/src/compiler.cpp
mitrokosta: Can I use storm::iEquals? We only need a function.

espkk: Sure, if it fits. But in that case, I believe you'll have to specify an iterable wrapper for Token.GetData(). Both range view and span should suffice...

UPD: nevermind, just convert to string; I missed the context

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -1347,27 +1339,22 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
...
             VMA *pClass;
 
             //-----------------------------------------------------
             // check if already loaded
-            bFound = false;
-            for (n = 0; n < LibriaryFuncs.GetClassesNum(); n++)
+            auto name = std::string_view(Token.GetData());
+            auto comparator = [name](const auto &library)
+            {
+                return storm::iEquals(library.name, name);
```
espkk, line 1351: This is fine now but still a bit weird. A simple range-based with boolean variable would look more descriptive...
But if you want to avoid it anyway in favor of functional style, there is a pattern using upper_bound with the desired binary predicate:
`if(std::ranges::upper_bound(LibriaryFuncs, name, storm, iEquals) != std::end(LibriaryFuncs)) {
  break;
}`
https://godbolt.org/z/c789hP8EG
Consider using it instead. 
Everything else is just fine 👍

espkk (changes requested):

espkk (approved):
Please, gnore my deleted comment, I looked through code superficially :/
It works and it uses the desired comparator, so let it be
Thanks for the PR!

espkk:
For future similar cases, while performing some refactoring we probably want to use std::set with iStrComparator. It makes code cleaner and reduces find complexity

PR #182, opened 2021-08-28
sea: improve MT for SSE_WaveXZBlock

espkk:
This basic MT wrapper for SSE_WaveXZBlock replaces the original one.
Initially, in low/mid-loaded sea scenes building wave blocks used to consume about 20% of CPU (single-core) time.
Timings should vary and depend on cores/hw-threads count; on my setup, I have ~x2 throughput for building waves with this PR; main thread CPU load decreased from ~20% to ~4-5%
I use it for a few months already, so it's kind of well-tested. But additional tests are always welcome.

### DIFF
--- a/src/libs/sea/src/SEA.cpp
+++ b/src/libs/sea/src/SEA.cpp
@@ -1,7 +1,9 @@
 #include "sea.h"
 
-#include "core.h"
+#include <algorithm>
+#include <execution>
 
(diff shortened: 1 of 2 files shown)


np-c0mp1ete (approved):
Nice that you removed a bunch of code. From the first glance the changes seem to be fine, thanks!

PR #183, opened 2021-08-29
location: backport camera MSGs from DMC

espkk:
This adds backport of 2 messages from the DMC engine: set radius and pos-to-pos view.
There is one more related to the deck camera but I have no idea of its purpose so I decided not to add it yet.

### DIFF
--- a/src/libs/location/src/LocationCamera.cpp
+++ b/src/libs/location/src/LocationCamera.cpp
@@ -365,6 +365,7 @@ uint64_t LocationCamera::ProcessMessage(MESSAGE &message)
         }
         return 1;
     case MSG_CAMERA_FOLLOW:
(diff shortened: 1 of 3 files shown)


espkk:
Please do approve but don't merge it yet. It has to be tested internally first

np-c0mp1ete (commented):
Why are we backporting messages from DMC? are they used in other games? shouldn't they be wrapped in `getTargetEngineVersion()` check?

espkk:
> 
> 
> Why are we backporting messages from DMC? are they used in other games? shouldn't they be wrapped in `getTargetEngineVersion()` check?

They are used in some mod. This actually the way how versioning had to be implemented :) This shouldn't anyhow affect other games

PR #184, opened 2021-08-30
Restore shadows for worldmap texts

mitrokosta:
Before/after picture for comparison. My commit reverses it.
image


np-c0mp1ete (approved):
Thanks for the fix!

PR #185, opened 2021-08-30
Fix worldmap zoom

mitrokosta:
After the SDL switch worldmap zoom stopped working because it relied on the absolute value of nMouseWheel which previously was a multiple of WHEEL_DELTA (120) and SDL output is +1 or -1. See more at https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel


espkk (approved):
Looks good, thanks!

PR #186, opened 2021-08-30
Fix TEHO-specific fencing stun chance

mitrokosta:
This is a change of a TEHO-specific feature, where characters would not always be stunned by an attack. It had a bug where the player character had semi-random stuns.

### DIFF
--- a/src/libs/location/src/Character.cpp
+++ b/src/libs/location/src/Character.cpp
@@ -4211,17 +4211,33 @@ void Character::UpdateAnimation()
                     fgtSetIndex = -1;
                     isFired = false;
                     break;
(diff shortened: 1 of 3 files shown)


espkk (changes requested):
According to the removed code, `stunChance` shall be initialized with `50`
Also, the original code has `IsPlayer()` check, why was it removed?

mitrokosta:
This behavior actually differs a lot across different versions. In CoAS only players get 50% of hits, in TEHO everyone is supposed to have their own stun chance determined by scripts, not being player. And earlier I think everyone had 100% stun chance. I'll fix this, will make a differentiation based on game version.

espkk (approved):
Looks good, thanks!

PR #188, opened 2021-08-31
engine: remove floating point precision format when converting to string

espkk:
gcvt was replaced with fmt::format with the same precision.
fmt::format follows sprintf semantic; it cannot match gcvt behaviour without manual transformation.
Also, I see no point in clipping float with a buffer size of `5` in general.
So, I removed precision in order to remove surplus 'width' precision.
Let's see if it will bloat something due to potentially long mantissa

### DIFF
--- a/src/apps/ENGINE/src/data.cpp
+++ b/src/apps/ENGINE/src/data.cpp
@@ -1675,7 +1675,7 @@ bool DATA::Plus(DATA *pV)
             Set(sValue + std::to_string(pV->lValue));
             break;
         case VAR_FLOAT:
-            Set(sValue + fmt::format("{:.5f}", pV->fValue));
+            Set(sValue + fmt::format("{}", pV->fValue));
             break;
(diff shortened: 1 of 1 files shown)


np-c0mp1ete (commented):
I guess it will lead to some UI bugs, not sure it's worth changing the current behavior, up to you.

espkk:
> I guess it will lead to some UI bugs, not sure it's worth changing the current behavior, up to you.

This was requested by @mitrokosta 
I am open to your better ideas :)

mitrokosta:
> I guess it will lead to some UI bugs, not sure it's worth changing the current behavior, up to you.

Without this PR there will be UI bugs such as this
image

np-c0mp1ete:
> Without this PR there will be UI bugs such as this
> image

I see, not ideal :) I guess, it makes sense to merge then.

np-c0mp1ete (approved):

PR #189, opened 2021-08-31
weather: fix stars existance check

espkk:
Should fix incorrect draw when stars are disabled for the current weather
thanks to @mitrokosta for localizing the issue


np-c0mp1ete (approved):
Looks good, thanks!

PR #190, opened 2021-09-01
xinterface: fix CXI_CHECKBUTTONS without individual pos

espkk:


np-c0mp1ete (dismissed):
hm, I can't figure out if the behavior has changed. obviously, you wouldn't have opened a PR if it stayed the same :)
so, I guess, it's just me not noticing it.

espkk:
> 
> 
> hm, I can't figure out if the behavior has changed. obviously, you wouldn't have opened a PR if it stayed the same :)
> so, I guess, it's just me not noticing it.

this was introduced by some other fix; fX and fY should persist through loop

PR #191, opened 2021-09-01
engine: empty strings allocated on pseudo-stack

espkk:
This is kinda weird but mirrors the original behavior that was changed in #142


np-c0mp1ete (approved):
thanks!

PR #192, opened 2021-09-01
engine: make token use utf8_inc

espkk:
Protects from potential collision


np-c0mp1ete (approved):
Hopefully, we don't have non-ascii characters in scripts, but indeed better to be on the safe side.

espkk:
> 
> 
> Hopefully, we don't have non-ascii characters in scripts, but indeed better to be on the safe side.

Any Russian symbol is already a non-ascii char :) This was an attempt to fix broken line counting inside the compiler; it did not help but still makes sense to be used

PR #194, opened 2021-09-02
window,engine: (!) rework window creation

espkk:
This is meant to be a basement for future features.
Basically, two things are planned: an ability to manage the main window from scripts (a basic example is shown in the snippet below) and to have entity-level access to window handles.

❗❗❗ This requires changes from the scripts' side and therefore shall be merged along with corresponding changes to the script codebases.

Example of integration:
```
void CreateMainWindow() {
	object MainWindow;
	MainWindow.title = VERSION_NUMBER1;
	CreateEntity(&MainWindow, "MainWindow");
}

void Main()
{
	CreateMainWindow();
...
```

### DIFF
--- a/src/apps/ENGINE/CMakeLists.txt
+++ b/src/apps/ENGINE/CMakeLists.txt
@@ -7,7 +7,6 @@ STORM_SETUP(
     common
     diagnostics
     util
-    window
 
     # VM
     animals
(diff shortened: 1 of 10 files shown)


espkk:
to be reimplemented as a service

PR #208, opened 2021-09-05
Fix keycode mapping for ALT key

mitrokosta:
Honestly I don't know what SDL_SCANCODE_MENU is, but it is not VK_MENU which is definitely ALT: https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes


espkk (approved):
Nice,thanks

PR #209, opened 2021-09-05
renderer: disable VSYNC by default

espkk:
- TODO: make it configurable

### DIFF
--- a/src/apps/ENGINE/src/Main.cpp
+++ b/src/apps/ENGINE/src/Main.cpp
@@ -58,7 +58,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
         return EXIT_SUCCESS;
     }
 
-    SDL_InitSubSystem(SDL_INIT_EVENTS | SDL_INIT_VIDEO | SDL_INIT_GAMECONTROLLER);
+    SDL_InitSubSystem(SDL_INIT_VIDEO | SDL_INIT_GAMECONTROLLER);
 
     // Init FS
     FILE_SERVICE File_Service;
--- a/src/libs/renderer/src/sdevice.cpp
+++ b/src/libs/renderer/src/sdevice.cpp
@@ -708,11 +708,8 @@ bool DX9RENDER::InitDevice(bool windowed, HWND _hwnd, long width, long height)
     // if(windowed) d3dpp.SwapEffect = D3DSWAPEFFECT_DISCARD;//FLIP;
     // else d3dpp.SwapEffect = D3DSWAPEFFECT_FLIP;
 
-    if (!windowed)
-    {
-        d3dpp.PresentationInterval = D3DPRESENT_INTERVAL_ONE;
-        // d3dpp.PresentationInterval = D3DPRESENT_INTERVAL_IMMEDIATE;
-    }
+    d3dpp.PresentationInterval = D3DPRESENT_INTERVAL_IMMEDIATE;
+    d3dpp.FullScreen_RefreshRateInHz = 0;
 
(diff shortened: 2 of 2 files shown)


src/apps/ENGINE/src/Main.cpp
```diff
@@ -58,7 +58,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
         return EXIT_SUCCESS;
     }
 
-    SDL_InitSubSystem(SDL_INIT_EVENTS | SDL_INIT_VIDEO | SDL_INIT_GAMECONTROLLER);
```
np-c0mp1ete, line 61: Event subsystem is initialized by default, right?

np-c0mp1ete (commented):
Why do we want to disable vsync?

src/apps/ENGINE/src/Main.cpp
espkk: Right, it is implied by SDL_INIT_VIDEO

espkk:
> 
> 
> Why do we want to disable vsync?

In general, it has a negative impact on fps. I know only a few people who actually use it.
But of course, this should be configurable as well as many other video settings. I plan to have a toml config in stash (documents) that may be configured manually until implemented in-game.
Also, HZ-based frame capping should be implemented (basically through setting specific maxFps automatically) for reasonable automatic frame capping without involving VSYNC

np-c0mp1ete:
> > Why do we want to disable vsync?
> 
> In general, it has a negative impact on fps. I know only a few people who actually use it.
> But of course, this should be configurable as well as many other video settings. I plan to have a toml config in stash (documents) that may be configured manually until implemented in-game.
> Also, HZ-based frame capping should be implemented (basically through setting specific maxFps automatically) for reasonable automatic frame capping without involving VSYNC

I mean, what's the point in unlimited fps if the monitor refresh rate is limited, not even saying about the human eye? :)
IMHO it will just introduce frame tearing and make rendering look buggy.
But again, feel free to merge, if you really want it.

espkk:
> 
> 
> > > Why do we want to disable vsync?
> > 
> > 
> > In general, it has a negative impact on fps. I know only a few people who actually use it.
> > But of course, this should be configurable as well as many other video settings. I plan to have a toml config in stash (documents) that may be configured manually until implemented in-game.
> > Also, HZ-based frame capping should be implemented (basically through setting specific maxFps automatically) for reasonable automatic frame capping without involving VSYNC
> 
> I mean, what's the point in unlimited fps if the monitor refresh rate is limited, not even saying about the human eye? :)
> IMHO it will just introduce frame tearing and make rendering look buggy.
> But again, feel free to merge, if you really want it.

VSYNC is just bad... it drastically increases latency and may screw up fps especially when the latter borderlines with HZ.
In my tests when I had fps drops from ~300to200 without VSYNC it resulted in ~144to110 drops with VSYNC enabled which is kind of distractive already.

As for me, I've never experienced screen tearing. But if it may appear we probably should try to involve triple buffering in windowed mode and gsync/adaptive sync or whatever it is called instead of vsync in fullscreen(do anyone indeed use fullscreen for non-fps games?).

Ok, maybe we should hold this until the optional setting is available, but I don't really want to introduce any new settings into engine.ini
So, consider this as blocked until I implement a new configuration system then

nbohr1more:
In native Doom 3, if you set r_swapInterval -1 this passes the associated swap attribute to the GL driver to enable "Adaptive Sync" which offers better performance while still performing vsync when possible

espkk:
> 
> 
> In native Doom 3, if you set r_swapInterval -1 this passes the associated swap attribute to the GL driver to enable "Adaptive Sync" which offers better performance while still performing vsync when possible

Yeah, that's what I meant 👍
But there are peculiarities depending on hardware and graphics API, so I cannot dive into it until higher prio tasks are done

PR #210, opened 2021-09-10
Replace conanfile.txt with conanfile.py; replace generator to support multi

espkk:
This basically aimed to support multi-config environments like Visual Studio.
Empowered with conanfile.py this PR also completely removes Steamworks dependency when building without steam and moves resource folder deploy into conanfile.py. Some dependencies were also updated/derived from the official artifactory.

What I expect to be broken atm:
CI (needs to adjust paths)
PROBABLY environments that don't use multi-config (e.g. ninja without multi-config enabled)

Could someone check the latter?


ugeen4:
just check this PR : checkout & remove .conan from home dir, remake cache (and update conan), compile (w/o steam), run TEHO. Don't see problems at all

np-c0mp1ete (commented):
Please check if everything still works when opening and building the project using the Visual Studio support for CMake.

espkk:
should be good now:
- build without sln is ok
- CI is ok
- correct fmod variant is used

Hammie (approved):
Also works for me!

PR #213, opened 2021-09-12
Re-enable interface scaling

Hammie:
Interface scaling is needed to run New Horizons properly.

Not sure why it was ever disabled, seems to work just fine.


espkk (approved):
Seems good to me. Worked fine when I introduced this, idk why it was commented out, likely because the original code (w/o scaling) was commented out in TEHO engine
@ugeen4, could you please check with TEHO if everything is alright

ugeen4:
All interfaces and fonts are correct after adding option bDynamicScaling = 1 in section [COMMON][PC_SCREEN] in file ini\interfaces\interfaces.ini
So, this PR may be approved and merged

PR #236, opened 2021-10-16
Fix setting button position from script

mitrokosta:
When you set button position from scripts, the coordinates range from 0 to 800, this is accounted for during initialization but not when changing position


np-c0mp1ete (approved):
Makes sense, thanks for the PR!

PR #237, opened 2021-10-18
Fix utf8 stash path

mitrokosta:
``const std::string toUtf8(u8str.begin(), u8str.end());`` mangles cyrillic characters in paths, results in crashes

### DIFF
--- a/src/libs/util/include/storm/fs.h
+++ b/src/libs/util/include/storm/fs.h
@@ -17,9 +17,7 @@ inline path GetStashPath()
         wchar_t *str = nullptr;
         SHGetKnownFolderPath(FOLDERID_Documents, KF_FLAG_SIMPLE_IDLIST, nullptr, &str);
         path = str;
-        const auto u8str = path.u8string();
-        const std::string toUtf8(u8str.begin(), u8str.end());
-        path = fs::path(toUtf8) / "My Games" / "Sea Dogs";
+        path = path / "My Games" / "Sea Dogs";
         CoTaskMemFree(str);
     }
     return path;


src/libs/util/include/storm/fs.h
```diff
@@ -17,9 +17,7 @@ inline path GetStashPath()
         wchar_t *str = nullptr;
         SHGetKnownFolderPath(FOLDERID_Documents, KF_FLAG_SIMPLE_IDLIST, nullptr, &str);
         path = str;
-        const auto u8str = path.u8string();
-        const std::string toUtf8(u8str.begin(), u8str.end());
```
espkk, line 21: Indeed, UTF-8 support still seems to be unstable. But we need to check everything very carefully to avoid introducing new issues.

I do not understand the point of `toUtf8`, it seems that `u8string` should be enough.
But anyway, this code does not "mangle" characters. The codepoint of Cyrillic char consists of two bytes, and therefore byte-array view becomes meaningless as Visual Studio only recognises ANSI/wchar_t.
It should be still a valid path in UTF-8.

Eventually, we want to achieve a transition to UTF-8 avoiding wchar_t. But if it helps to solve issues we can leave it that way.
But then we need to make sure that every part of code that indirectly gets this function's value expects wchar_t.

Briefly, I can remember `sentry-native` likely indeed expects native encoding. But we have to check and test(debug) that and all other places. I think there are only a few

espkk (changes requested):

src/libs/util/include/storm/fs.h
np-c0mp1ete: on a side note, to properly view utf8 strings in VS you need to add `s8` like explained [here](https://docs.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2019#BKMK_Visual_Studio_2012_format_specifiers). maybe that helps

espkk (approved):
Aight, everything seems to be working with the native encoding.
Thanks for the PR!

PR #238, opened 2021-10-19
Fix VK_SHIFT behaviour

mitrokosta:
VK_SHIFT means either left shift is pressed or right shift is pressed, my PR brings this back

### DIFF
--- a/src/libs/pcs_controls/src/pcs_controls.cpp
+++ b/src/libs/pcs_controls/src/pcs_controls.cpp
@@ -346,6 +346,10 @@ bool PCS_CONTROLS::GetControlState(long control_code, CONTROL_STATE &_state_stru
                 if (!input_->KeyboardKeyState(VK_MENU))
                     pressed = input_->KeyboardKeyState(system_code);
             }
+            else if (system_code == VK_SHIFT)
+            {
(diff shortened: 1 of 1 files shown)


espkk (approved):
Thanks for the PR!

`brings back` to what state? Is this related to SDL2 or not? 
Should we introduce VK_RSHIFT/VK_LSHIFT, or it will have a collision with legacy scripts?

If it brings vanilla behaviour and separating virtual codes is much effort that is good for me to merge it as is!

mitrokosta:
This change is related to the SDL switch, SDL does not appear to have a scancode analogy for VK_SHIFT.
In scripts for interface theoretically we can separate LSHIFT/RSHIFT:
```
CI_CreateAndSetControls( "", "ILeftShift", CI_GetKeyCode("VK_SHIFT"), 0, false );
CI_CreateAndSetControls( "", "IRightShift", CI_GetKeyCode("VK_SHIFT"), 0, false );
```
But in some places there is no easy way to do it
``CI_CreateAndSetControls( "PrimaryLand", "ChrRun", CI_GetKeyCode("VK_SHIFT"), USE_AXIS_AS_BUTTON, true );``

PR #239, opened 2021-10-19
Make location health bars configurable

mitrokosta:
By using ``MSG_LOCATION_VIEWSTATEBARS`` message we can hide or show health bars


espkk (approved):
👍

PR #241, opened 2021-10-23
Fix AI shooting through allies logic

mitrokosta:
Sometimes autoaim does not let you shoot enemies that are literally in front of you.

On this screenshot my ship can clearly shoot the galleon in the circle, but the allied warship to the left somehow does not let me.
image

### DIFF
--- a/src/libs/sea_ai/src/AIShip.cpp
+++ b/src/libs/sea_ai/src/AIShip.cpp
@@ -323,9 +323,9 @@ bool AIShip::isCanFire(const CVECTOR &vFirePos) const
                 continue;
             if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vFirePos))
                 return false;
-            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, v1))
+            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vOurPos + v1))
                 return false;
-            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, v2))
+            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vOurPos + v2))
                 return false;
         }
 
--- a/src/libs/sea_ai/src/AIShip.cpp
+++ b/src/libs/sea_ai/src/AIShip.cpp
@@ -304,30 +304,32 @@ void AIShip::CheckStartPosition() const
 
 bool AIShip::isCanFire(const CVECTOR &vFirePos) const
 {
-    CVECTOR v1, v2, vOurPos;
-    float fAng, fCos, fSin;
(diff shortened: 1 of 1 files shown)


src/libs/sea_ai/src/AIShip.cpp
```diff
@@ -305,17 +305,16 @@ void AIShip::CheckStartPosition() const
...
+    float fAng, fCos, fSin;
 
     vOurPos = GetPos();
 
     fAng = 8.0f * PI / 180.0f;
     fCos = cosf(fAng), fSin = sinf(fAng);
-    fDist = sqrtf(~(vOurPos - vFirePos));
 
-    v1 = v2 = fDist * !(vFirePos - vOurPos);
+    v1 = v2 = vFirePos - vOurPos;
```
espkk: makes sense

src/libs/sea_ai/src/AIShip.cpp
```diff
@@ -305,17 +305,16 @@ void AIShip::CheckStartPosition() const
...
     RotateAroundY(v1.x, v1.z, fCos, fSin);
-    RotateAroundY(v2.x, v2.z, -fCos, -fSin);
+    RotateAroundY(v2.x, v2.z, fCos, -fSin);
```
espkk, line 317: makes sense as well

src/libs/sea_ai/src/AIShip.cpp
```diff
@@ -324,9 +323,9 @@ bool AIShip::isCanFire(const CVECTOR &vFirePos) const
                 continue;
             if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vFirePos))
                 return false;
-            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, v1))
+            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vOurPos + v1))
```
espkk: I believe it should be also divided by fCos to be more precise

src/libs/sea_ai/src/AIShip.cpp
```diff
@@ -324,9 +323,9 @@ bool AIShip::isCanFire(const CVECTOR &vFirePos) const
...
                 return false;
-            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, v2))
+            if (AIShips[i]->GetTouchController()->isCollision2D(vOurPos, vOurPos + v2))
```
espkk: same here

espkk (changes requested):
Good find, thanks for the PR! 
I'm wondering if this is original behaviour or not...
Please, check the `cos` thing and also do Resharper 1-click-refactoring things for declarations and assignments and also for the loop

espkk (approved):
Looks good, thanks

PR #245, opened 2021-10-26
Add dynamic lighting for island

mitrokosta:
Originally proposed by @Mett95 and sent to me by @ugeen4

If you have questions concerning this PR, please address these people


espkk:
Thanks. Is this code already refined? The original one had some unwanted side effects that I wanted to avoid...

ugeen4:
Yes, this code checked from my side - no unwanted effects at all

espkk (approved):
good, then it can be merged

PR #247, opened 2021-10-27
gamesave convert

mitrokosta:


espkk (changes requested):
can we reject this after merging #262 #263?

mitrokosta:
Sure thing

PR #250, opened 2021-10-28
Fix steam achievements

mitrokosta:
Steam achievements were broken because achievement functions used pointers to stack data erased by subsequent stack push


espkk (approved):
Ok for me
maybe @ugeen4 wants to take a look

ugeen4 (approved):

PR #253, opened 2021-10-31
battle_shipsign: make picture count configurable

Artess999:
Currently I'm working on AoP Caribbean Tales open source SE conversion
It has different battle interface ship icon pictureWidth and pictureHeight than CoAS / TEHO

Also will be useful for mods like CSP

Снимок экрана 2021-11-01 в 02 20 15

Снимок экрана 2021-11-01 в 02 19 55

@mitrokosta @ugeen4 Could you test that nothing breaks on TEHO?


Hammie (approved):

mitrokosta (approved):
Does not appear to break TEHO

PR #254, opened 2021-11-02
Fix GetSymbol function for UTF-8

mitrokosta:
Used to return only one codepoint


src/apps/ENGINE/src/internal_functions.cpp
```diff
@@ -432,7 +432,7 @@ DATA *COMPILER::BC_CallIntFunction(uint32_t func_code, DATA *&pVResult, uint32_t
 
     long slen, slen2;
     char sVarName[64];
-    char sBuff2[2];
+    char8_t utf8_character[5];
```
np-c0mp1ete: maybe better to use std::string here? that would simplify the code a tiny bit, I think

np-c0mp1ete (approved):

PR #257, opened 2021-11-05
Add XI_IsVirtualKeyPressed internal function

mitrokosta:
XI_IsKeyPressed is very limiting (only 3 keys available), so I added this function that takes an arbitrary virtual key. Originally requested by Hokkins.


espkk (approved):
Good, thanks

PR #258, opened 2021-11-07
[meta,diagnostics,engine] Add basic watermarking

espkk:
Introduce basic build watermarking which gets printed in logs/can be sorted by with sentry.


espkk:
TODO: pass hash from CI job

PR #260, opened 2021-11-15
docs: add missing foam editor doc

Artess999:


np-c0mp1ete (approved):
Unfortunate to have it only in Russian, but still better than no docs at all.

PR #261, opened 2021-11-15
Fix enter and free camera controls

mitrokosta:


espkk (approved):
Looks good to me, thanks

PR #262, opened 2021-11-16
Save tweaks for compatibility with older saves

mitrokosta:


src/libs/sea_ai/src/AIBalls.cpp
```diff
@@ -434,6 +428,27 @@ uint64_t AIBalls::ProcessMessage(MESSAGE &message)
     return 0;
 }
 
+void BALL_PARAMS::Save(CSaveLoad *pSL)
```
espkk, line 431: @np-c0mp1ete why this 'Save' is needed? Cannot we just define pointer as an uint64_t instead?

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -6120,6 +6120,7 @@ char *COMPILER::ReadString()
 
     char *pBuffer = new char[n];
     ReadData(pBuffer, n);
+    //Assert(utf8::IsValidUtf8(pBuffer));
```
espkk: Should we log this behaviour? It may help detecting non-converted files

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -6593,7 +6613,7 @@ bool COMPILER::LoadState(std::fstream &fileS)
         pString = ReadString();
         if (pString)
         {
-            Assert(utf8::IsValidUtf8(pString));
+            //Assert(utf8::IsValidUtf8(pString));
```
espkk: Should we log this behaviour? It may help detecting non-converted files

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -6626,7 +6646,6 @@ bool COMPILER::LoadState(std::fstream &fileS)
             SetError("missing variable name");
             return false;
         }
-        Assert(utf8::IsValidUtf8(pString));
```
espkk, line 6629: Should we log this behaviour? It may help detecting non-converted files

src/apps/ENGINE/src/compiler.cpp
```diff
@@ -6205,6 +6206,25 @@ bool COMPILER::ReadVariable(char *name, /* DWORD code,*/ bool bDim, uint32_t a_i
         // load array elements
         for (uint32_t n = 0; n < nElementsNum; n++)
         {
+            if (bSkipVariable)
```
espkk, line 6212: Can you provide some context why this is required?

src/apps/ENGINE/src/compiler.cpp
np-c0mp1ete: sure, replaced with spdlog::warn

src/apps/ENGINE/src/compiler.cpp
np-c0mp1ete: `bSkipVariable` is true when there is no such variable in the scripts. It might happen when you make a save, remove a variable from the scripts and then try to load the save.
This code skips this variable data when reading the save file.

src/apps/ENGINE/src/compiler.cpp
np-c0mp1ete: since now there is a check in the `ReadString` function, no need to have the same check here. I removed it.

src/apps/ENGINE/src/compiler.cpp
np-c0mp1ete: since now there is a check in the `ReadString` function, no need to have the same check here. I removed it.

src/libs/sea_ai/src/AIBalls.cpp
np-c0mp1ete: it is needed because I changed the type of `sBallEvent` from `char[]` to `std::string`. Also, this way it's explicit what we want to serialize and how.

espkk (approved):
Looks good to me, thanks!

PR #263, opened 2021-11-16
Add python converter for saves made on storm 2.8

mitrokosta:


espkk (dismissed):
Looks good to me!
@np-c0mp1ete, can you confirm this is ready for merge?

np-c0mp1ete (approved):
Should be good now, merging it.

PR #269, opened 2021-11-24
Revert interface scaling changes when dynamic scaling is set to 1

mitrokosta:
#213 broke interfaces for 1280x1024 resoultion

### DIFF
--- a/src/libs/xinterface/src/xinterface.cpp
+++ b/src/libs/xinterface/src/xinterface.cpp
@@ -1083,21 +1083,9 @@ void XINTERFACE::LoadIni()
     sprintf_s(section, "COMMON");
 
     // set screen parameters
-    const auto &canvas_size = core.GetScreenSize();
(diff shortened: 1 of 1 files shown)


espkk:
Could you explain the idea? How the interfaces will be scaled without that code?

mitrokosta:
> Could you explain the idea? How the interfaces will be scaled without that code?

I honestly have no idea, but it worked before with all this code commented out. Hammie needs only this if.

Hammie:
I checked the game scripts and I don't see `bDynamicScaling` being used anywhere. Don't even see it in TEHO, so also not sure how it could be causing problems.

ugeen4:
seems ok for TEHO

ugeen4 (approved):

np-c0mp1ete (approved):
It seems it's fine for everybody to merge it, thanks for the PR!

PR #270, opened 2021-11-25
blender-gm-export: remove BSP and add triangulation option

Artess999:


np-c0mp1ete (approved):
Thanks for supporting the tool!

Could you explain why BSP is not needed anymore please?

Artess999:
> Thanks for supporting the tool!
> 
> Could you explain why BSP is not needed anymore please?

Снимок экрана 2021-11-25 в 18 00 34

Also seems that BSP in Blender could give wrong results (at least some users says so)
I think that this happens because of heavy recursion use, which is not welcomed in python (not sure thought)

Artess999:
Also bsp-patcher is better because basically it is wrapper with gm read/write and for BSP calculation it is using slightly modified Nick Chirkov's original BSP code

PR #276, opened 2021-12-10
Refactor options parsing to remove hard-coded character limit

Hammie:
Refactored the options parsing code to fix an issue with a hard-coded character limit (of 512 characters). New Horizons has some option files with longer values which was causing undefined behavior (stack corruption).

Also added some tests to verify everything works.

In terms of performance it introduces a few extra memory allocations (strings), these could be avoided by making the `ATTRIBUTES` class work with `std::string_view`s. But I don't think this will currently affect game performance noticeably.

Any feedback or suggestions welcome!

### DIFF
--- a/src/libs/xinterface/include/storm/xinterface/options_parser.hpp
+++ b/src/libs/xinterface/include/storm/xinterface/options_parser.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#include "attributes.h"
+
+#include <string>
+
+namespace storm {
+
+void removeCarriageReturn(std::string &str);
+
+void parseOptions(const std::string_view &str, ATTRIBUTES &attribute);
+
+} // namespace storm
--- a/src/libs/xinterface/CMakeLists.txt
+++ b/src/libs/xinterface/CMakeLists.txt
@@ -1,5 +1,6 @@
 STORM_SETUP(
     TARGET_NAME xinterface
     TYPE storm_module
-    DEPENDENCIES common ddraw amstrmid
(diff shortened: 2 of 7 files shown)


src/libs/xinterface/include/storm/xinterface/options_parser.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "attributes.h"
+
+#include <string>
+
+namespace storm {
+
+namespace detail {
```
espkk: To my knowledge, `detail` meant to hide impl details of the code inlined into headers (templates or h-only libraries). Here it is used as a self-sufficient code, therefore I would rename that namespace to smth like`util`.
I checked our `util` lib and there is also `detail` instead of `util` that I would also rename...

But naming is not a big problem, and everything else seems just good to me. So anyway, I just want to know your opinion, @Hammie

np-c0mp1ete (dismissed):
Looks good to me, thanks!

src/libs/xinterface/include/storm/xinterface/options_parser.hpp
Hammie: I like to put things in `detail` that should not be used in user code, but for some reason need to be declared/defined in a header. In this case I forgot to remove it as `removeCarriageReturn` is being used (and should be), so I suggest just removing `detail` here and leaving it in the `storm` namespace?

I would leave the ones in `util` though, as those are not intended to be used for anything else.

espkk (approved):
👍

PR #277, opened 2021-12-12
Procedural head look

dimogrudev:
Engine improvements
https://youtu.be/Y9dhvzu7Y0Y

### DIFF
--- a/src/libs/location/src/Character.cpp
+++ b/src/libs/location/src/Character.cpp
@@ -611,7 +611,7 @@ Character::Character()
 
     // Procedural head look
     curHeadLookState = headlook_default;
-    headLookChrTarget = nullptr;
+    headLookChrTarget = {};
     headLookPointTarget = 0.0f;
     curHeadAX = 0.0f;
     curHeadAY = 0.0f;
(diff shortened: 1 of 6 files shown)


src/libs/location/src/Character.cpp
```diff
@@ -796,6 +803,29 @@ uint64_t Character::ProcessMessage(MESSAGE &message)
...
+        curHeadLookState = headlook_default;
+        return 1;
+    case MSG_CHARACTER_HEADLOOK_CAMERA:
+        curHeadLookState = headlook_camera;
+        return 1;
+    case MSG_CHARACTER_HEADLOOK_CHARACTER: {
+        curHeadLookState = headlook_character;
+
+        const entid_t chr = message.EntityID();
+        headLookChrTarget = static_cast<Character *>(EntityManager::GetEntityPointer(chr));
```
espkk: Don't store ptr - store entid_t instead. Get ptr each time you need it and test against nullptr

espkk (changes requested):
Thanks for the PR!
Apart from the mentioned point everything looks good to me
Maybe someone can test it as well

src/libs/location/src/Character.cpp
dimogrudev: Thanks for the feedback!
I've fixed it.

PR #278, opened 2021-12-12
Procedural head look

dimogrudev:
https://youtu.be/Y9dhvzu7Y0Y
I had to re-open the PR because I've closed it by mistake, sorry

### DIFF
--- a/src/libs/location/src/Character.cpp
+++ b/src/libs/location/src/Character.cpp
@@ -610,7 +610,7 @@ Character::Character()
     m_nHandLightID = -1;
 
     // Procedural head look
-    curHeadLookState = headlook_default;
+    curHeadLookState = HeadLookState::dflt;
     headLookChrTarget = {};
     headLookPointTarget = 0.0f;
     curHeadAX = 0.0f;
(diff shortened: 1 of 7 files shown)


Artess999:
Have you checked on other skeletons except man.an? I'm not sure that neck bone idx is always 16
Maybe it could be moved to scripts and if it is not set from scripts it'll be 16?

dimogrudev:
> Have you checked on other skeletons except man.an? I'm not sure that neck bone idx is always 16 Maybe it could be moved to scripts and if it is not set from scripts it'll be 16?

I've added an option in .ani file. Now you can change head bone index by
`data = "Head bone index", "16"`

Artess999:
> > Have you checked on other skeletons except man.an? I'm not sure that neck bone idx is always 16 Maybe it could be moved to scripts and if it is not set from scripts it'll be 16?
> 
> I've added an option in .ani file. Now you can change head bone index by `data = "Head bone index", "16"`

Thanks!

src/libs/location/src/Character.cpp
```diff
@@ -2450,6 +2477,103 @@ void Character::Update(float dltTime)
...
+        {
+        case headlook_default:
+            break;
+
+        case headlook_camera:
+            destHeadAX = camAng.x - 0.2f;
+            destHeadAY = atan2(sin(ay - camAng.y), cos(ay - camAng.y));
+            break;
+
+        case headlook_character:
```
espkk: This is evil and normally we want to get rid of such code: instead, additional `if` or lambda for shared code should've been used. 
But it is FYI, I do not insist on rewriting this part.

src/libs/location/src/Character.cpp
```diff
@@ -2450,6 +2477,103 @@ void Character::Update(float dltTime)
...
+        case headlook_point: {
+            CVECTOR targetPos = 0.0f;
+            if (curHeadLookState == headlook_character)
+            {
+                auto *targetChrPtr = static_cast<Character *>(EntityManager::GetEntityPointer(headLookChrTarget));
+                if (targetChrPtr)
+                {
```
espkk, line 2511: nullptr means the entity was deleted. It would make sense to change `curHeadLookState` to `headlook_default` and break

src/libs/location/src/Character.h
```diff
@@ -251,6 +251,15 @@ class Character : public Entity
     //--------------------------------------------------------------------------------------------
     // Character animation
     //--------------------------------------------------------------------------------------------
+  protected:
+    // Procedural head look
+    enum HeadLookState
```
espkk: Could you use `enum class` instead, please? There is a lot of nasty legacy things, but for the new code, we prefer to avoid bad practices.

espkk (changes requested):
Thanks for the PR again 👍 
Please check the additional comments and I think it is to be merged right after.

dimogrudev:
Thank you @espkk

espkk (approved):
thanks, all's good now! 👍

PR #280, opened 2021-12-14
[renderer] fix DrawRects draw size

espkk:


Hammie (approved):
It's a mind-bending piece of code, but I'm pretty sure this does indeed fix an issue when `dwRectsNum` > `rectsVBuffer_SizeInRects`

PR #281, opened 2021-12-20
SoundService: Audio Pitch Controller

dimogrudev:
A good feature for fast/slow motion effect
https://youtu.be/qvSbro32CLY


espkk (approved):
I cannot imagine the situation when we need to get the pitch from outside :)
But anyway, looks good to me, thanks for the PR!

PR #282, opened 2021-12-26
Fix crashpad initialization when built without crash reports

mitrokosta:
When you build without crash reports, crashpad_handler.exe has to be present anyway, or else you get the warning message box every time

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -47,8 +47,8 @@ class StormEngine(ConanFile):
         else:
             self.__intall_lib("fmod.dll")
 
+        self.__install_bin("crashpad_handler.exe")
         if self.options.crash_reports:
-            self.__install_bin("crashpad_handler.exe")
             self.__install_bin("7za.exe")
 
         if self.options.steam:


espkk (approved):
Thanks for the PR!
But shouldn't we disable sentry initialization instead if crash reports are disabled?
Initially, the idea was to disable only report uploading while still collecting some events.
But since events are not used I'm wondering if it still makes any sense...

P.S. external crash reporter is requested upon sentry_options_new invocation

mitrokosta:
> shouldn't we disable sentry initialization instead if crash reports are disabled? Initially, the idea was to disable only report uploading while still collecting some events. But since events are not used I'm wondering if it still makes any sense...

It does make sense to enable sentry initialization. Here's a use case that I just tested. I don't want steam or crash reports in my test version, but I do want error.log flushed when the game crashes, especially at startup if there are compile errors. If I compile with crash reports disabled and don't include crashpad handler in my distribution, this ruins sentry initialization and when the game crashes, logs are not flushed and it is a lot harder to determine the source of error without resorting to debugging from VS. If I do include crashpad handler, even with reports disabled, this gets me rid of the message box and flushes logs in the case of a crash.

espkk:
Well, I'm thinking about overhauling that diag code anyway, so let it be so for now, and then we'll see

PR #284, opened 2021-12-29
Remove dead and duplicated code

q4a:
There was 2 identical files (except comments in it):
`/src/libs/location/src/ptc.h`
`/src/libs/worldmap/src/ptc.h`
Looks like `location/src/ptc.h` best one and should be in `Common\include`

Class `CXPICTURES` has only header and did not used anywhere.


espkk (approved):
Looks good, thank you!

PR #285, opened 2021-12-30
Fix comparing strings with raw string literals and count parameter

Hammie:
Fixes case-insensitive comparisons for raw string literals: `const char*` and `const char[]`.

Also adds an option to only compare `count` number of characters, for partial string comparisons.

### DIFF
--- a/src/libs/util/include/storm/string_compare.hpp
+++ b/src/libs/util/include/storm/string_compare.hpp
@@ -34,28 +34,72 @@ struct is_iless_eq
 
 } // namespace detail
 
+template <typename Range1T, typename Range2T = Range1T>
+bool iEquals(const Range1T &first, const Range2T &second, const size_t count)
+{
+    detail::is_iequal comp;
+
+    const auto &first_normalized = std::is_pointer<Range1T>::value ? std::string_view(first) : first;
+    const auto &second_normalized = std::is_pointer<Range1T>::value ? std::string_view(second) : second;
+
+    const auto first_begin = std::begin(first_normalized);
+    const auto second_begin = std::begin(second_normalized);
+
+    const auto first_end = std::end(first_normalized);
+    const auto second_end = std::end(second_normalized);
+
+    const auto first_length = std::distance(first_begin, first_end) - 1;
+    const auto second_length = std::distance(first_begin, first_end) - 1;
+    if (first_length < count || second_length < count)
+    {
+        if (first_length != second_length)
+        {
+            return false;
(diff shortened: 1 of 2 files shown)


q4a:
Sorry for delay. I found error, but don't understand where it was. This example returns true, but should be false:
`std::cout << "comp08=0: " << iEquals(".txt1", ".txt2", 5) << "\n";`
I think that better to write:
```
template <typename Range1T, typename Range2T = Range1T>
bool iEquals(const Range1T &first, const Range2T &second, const size_t count)
{
    detail::is_iequal comp;

    const auto &first_normalized = std::is_pointer<Range1T>::value ? std::string_view(first) : first;
    const auto &second_normalized = std::is_pointer<Range1T>::value ? std::string_view(second) : second;

    const auto first_begin = std::begin(first_normalized);
    const auto second_begin = std::begin(second_normalized);

    const auto first_end = std::end(first_normalized);
    const auto second_end = std::end(second_normalized);

    const auto first_length = std::distance(first_begin, first_end);
    const auto second_length = std::distance(second_begin, second_end);
    if (first_length < count || second_length < count)
    {
        if (first_length != second_length)
        {
            return false;
        }
        else
        {
            return std::equal(first_begin, first_end, second_begin, second_end, comp);
        }
    }
    else
    {
        return std::equal(first_begin, first_begin + count, second_begin, second_begin + count, comp);
    }
}
```

src/libs/util/include/storm/string_compare.hpp
```diff
@@ -34,28 +34,72 @@ struct is_iless_eq
...
+    const auto &second_normalized = std::is_pointer<Range1T>::value ? std::string_view(second) : second;
+
+    const auto first_begin = std::begin(first_normalized);
+    const auto second_begin = std::begin(second_normalized);
+
+    const auto first_end = std::end(first_normalized);
+    const auto second_end = std::end(second_normalized);
+
+    const auto first_length = std::distance(first_begin, first_end) - 1;
+    const auto second_length = std::distance(first_begin, first_end) - 1;
```
q4a: here is mistake

q4a (changes requested):

src/libs/util/include/storm/string_compare.hpp
Hammie: I think I fixed it for real this time

q4a (approved):
Looks good, thank you!

espkk (approved):

PR #287, opened 2021-12-31
Change file names according code style

q4a:
I did task: https://github.com/storm-devs/storm-engine/projects/1#card-57680541
Used this:
https://google.github.io/styleguide/cppguide.html#File_Names

Deleted 5 unused files:
src/libs/Common/include/safequeue.h (115 lines)
src/libs/particles/src/ICommon/BillBoard.h (42 lines)
src/libs/particles/src/ICommon/EditorTypes.h (22 lines)
src/libs/sea_operator/src/TFIFOBuffer.h (70 lines)
src/libs/sea/src/sea_struct.h (45 lines)

Removed 4 commented out include:
```
//#include "..\file.h"
//#include "AITemplate.h"
//#include "TFIFOBuffer.h"
//#include "TSharks.h"
```
Line changes check:
126,437 additions - 126,735 deletions = -298 = - (115 + 42 + 22 + 70 + 45 +4)


espkk:
Great job! This looks good to me. I was able to fetch fresh case-sensitive repo and build.
However, files from `src\libs\shared_headers\include\shared\` are used directly in the scripts. 
That means we should either don't touch them, or to change paths in scripts as well.
Personally, I like the second approach. They are used in the engine as well, and engine code quality should be in the first place. Also, the scripts that are not adjusted for the engine yet need these paths to be modified anyway, because of the newly added prefix.

So, let's wait for the others to comment. If there are no objections, let's prepare patches to the scripts before merging this.

espkk:
@ugeen4 , are you ok with this? a few path changes in the scripts are required 
If no objections, I'll do rebase and merge

espkk (dismissed):
After closer look it seems that only 1 header is affected (on windows):
`SeaPeople.h` -> `sea_people.h` in `sea.c`
I'm marking this as script-breaking and preparing for merge

espkk (approved):

PR #288, opened 2022-01-04
Optimizations (grass fix here)

espkk:
a bunch of profiling/optimization stuff
most significant part is grass optimizations for sea. no more lags


ugeen4 (approved):
Great ! Thank you for PR !

PR #293, opened 2022-01-05
[location] fix fader PICTURE0 draw

espkk:


espkk:
seems to be OK with TEHO

ugeen4 (approved):

PR #302, opened 2022-01-08
Replace long with int32_t

q4a:
Replace:
long => int32_t
unsigned long => uint32_t
unsigned long int => uint32_t
unsigned long long => uint64_t

This will increase the portability of the code.


q4a:
I can add another change to this PR: replacing GetLong* functions to GetInt*
Like this one: https://github.com/q4a/storm-engine/commit/e5c25fe725f571e9419705f554808d727ff541e3

espkk (changes requested):
`atol` won't work for parsing 32-bit unsigned integers :(
we should either bring back `atoll` or find a better way

espkk (approved):
Looks good, no regressions observed
Thanks!

PR #310, opened 2022-01-11
replace MSVC-only *vsnprintf_s, _snprintf, _ultoa_s, _ltoa_s

q4a:


espkk:
Looks good, thanks!

`gcvt` will generate a warning with msvc though. since it's only one place where it's used probably we can find a way to replace it. Honestly, I think I already tried to do that, but ended up restoring the original call :)

the only thing I'm concerned, did you check that `sprintf` generates exactly the same output with `_ltoa_s`?

q4a:
> the only thing I'm concerned, did you check that `sprintf` generates exactly the same output with `_ltoa_s`?

Yes.

espkk (approved):

PR #315, opened 2022-01-12
fix ship lights on/off during day/night cycle without location reloading

ugeen4:
Also remove some dead code & fix ship fake fire

### DIFF
--- a/src/libs/ship/src/ship_lights.cpp
+++ b/src/libs/ship/src/ship_lights.cpp
@@ -147,7 +147,7 @@ void ShipLights::AddDynamicLights(VAI_OBJBASE *pObject, const CVECTOR &vPos)
     light.pObject = pObject;
     light.vPos = vPos;
     light.bOff = false;
-	light.bTmpOff = false;
+    light.bLightOff = false;
     light.bBrokenTimeOff = false;
     light.fCurTime = 0.0;
     light.fTotalTime = pLT->fLifeTime;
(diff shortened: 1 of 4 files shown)


src/libs/ship/src/ship_lights.cpp
```diff
@@ -655,13 +643,26 @@ void ShipLights::Realize(uint32_t dwDeltaTime)
...
+void ShipLights::ResetLights(VAI_OBJBASE *pObject, bool bLight)
+{
+	for (uint32_t i = 0; i < aLights.size(); i++)
+	{
+		ShipLight &L = aLights[i];
+		if (L.bOff)
+            continue;
+		if(bLight) 	L.bTmpOff = false;
+		else		L.bTmpOff = true;
+	}
```
espkk:
```suggestion
    for (auto &light : aLights)
    {
        if (!light.bOff)
        {
            light.bTmpOff = !bLight;
        }
    }		
```
would be more expressive

src/libs/ship/src/ship_lights.h
```diff
@@ -82,6 +82,7 @@ class ShipLights : public IShipLights
         bool bVisible;
         float fTotalBrokenTime, fBrokenTime;
         bool bOff, bBrokenTimeOff;
+		bool bTmpOff;
```
espkk: The naming is counter-intuitive. could you please rename it to something meaningful...

espkk (changes requested):
Thank you!
Seems good, but please check the comments below + please fix the formatting, if you look at the diff most of the new code has a different indent (most likely due to tabs instead of spaces)

espkk (approved):
Great, thanks!

NB: MSG_SHIP_LIGHTSRESET message was changed

PR #316, opened 2022-01-12
Changed Ship Speed Influence In Cannons Aiming

LipsarJ:
Hello. I noticed, that balls, sometimes are flying too far in forward of enemy ship and I am missing shots. So I redused ship's speed influence in cannons aiming. I've tested it on Kaleuche as the fastest ship and successfully destroy this ship. I saw good aiming, with almost zero misses on highest speed


espkk (changes requested):
That makes no sense before investigating the issue since the formula seems to be correct.
Instead, we need to figure out what parameter is wrong here and why

LipsarJ:
> That makes no sense before investigating the issue since the formula seems to be correct. Instead, we need to figure out what parameter is wrong here and why

All parameters, are OK by theirself. But together they are wrong. Ships are not so fast, to put their full speed in formula. And if it  works correctly, balls should fall in centre of ship, but it isnt so. If u play game u see it.

PR #317, opened 2022-01-13
Ship Lines Addition

LipsarJ:


espkk:
thanks for the PR! I propose to this approach instead: https://github.com/storm-devs/storm-engine/pull/352 I will provide a snippet for default pos calculation. It has to be adjusted for different formations like this one

PR #318, opened 2022-01-15
use va_list instead of (char *) in vsnprintf

q4a:


Hammie (approved):
Thank you!

espkk (approved):
All good, thanks!

PR #308, opened 2022-01-09
replace _stricmp with storm::iEquals

q4a:
same for _strcmpi

PR #320, opened 2022-01-16
replace _strnicmp with storm::iEquals

q4a:
It should be in this PR: https://github.com/storm-devs/storm-engine/pull/308
But I forgot to add it)


espkk (approved):
Thanks!

PR #321, opened 2022-01-16
Ballistics prediction improvement and debug draw for sea AI

espkk:
Prediction improved (actually fixed) by taking into account `AIBalls.SpeedMultiply` and fire time; improved height handling.

I also suggest rewriting SLIB `CANNON_GET_FIRE_HEIGHT` script event in a way it:
1) does not depend on distance since it should be an additional height relative to a target model
2) has a stronger dispersion

For debug draw press `H` with debug keys enabled.


Hammie (approved):
Tested it with TEHO, and seems to work well for me

ugeen4 (approved):

PR #324, opened 2022-01-17
[core] Integrate mimalloc override allocators

espkk:
Initial mimalloc integration.
Tested for a long time with no issues, yet they may appear with some DLLs on some particular windows versions.

### DIFF
--- a/src/apps/engine/CMakeLists.txt
+++ b/src/apps/engine/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)
+
 STORM_SETUP(
     TARGET_NAME engine
     TYPE executable
--- a/src/apps/engine/CMakeLists.txt
+++ b/src/apps/engine/CMakeLists.txt
@@ -1,8 +1,11 @@
 STORM_SETUP(
     TARGET_NAME engine
     TYPE executable
-    DEPENDENCIES     
-	
+    DEPENDENCIES
+    
+    # this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
+    $<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>
+
     # internal
     common
     diagnostics
@@ -50,9 +53,10 @@ STORM_SETUP(
     weather
     worldmap
     xinterface
-	
+
     # external
     fast_float
+    mimalloc
     sentry-native
     sdl
     zlib
--- a/src/apps/engine/CMakeLists.txt
+++ b/src/apps/engine/CMakeLists.txt
@@ -1,12 +1,17 @@
+# this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
 include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)
+if (WIN32)
+  set(MIMALLOC_DEP "$<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>")
+else()
+  set(MIMALLOC_DEP "$<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/libmimalloc.so,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/libmimalloc-debug.so>")
+endif()
 
 STORM_SETUP(
     TARGET_NAME engine
     TYPE executable
     DEPENDENCIES
-    
-    # this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
-    $<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>
+
+    ${MIMALLOC_DEP}
 
     # internal
     common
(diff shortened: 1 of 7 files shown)


ugeen4:
Tested. Don't see any problems with this PR !
 Why is the stable release (mimalloc 1.7.3) not being used?

q4a:
Can you wait for 1-2 days before merge? I would like to test this changes on Linux.

espkk:
@ugeen4, I tested both for quite a lot of time.
2.* is quite stable. the reason it is in beta I think is that it is not as well tested as 1.*
Considering the 2.* main difference is the internal resource management improvement which results in better performance, I decided to go with 2.* and see if it has some issues.
And I'm happy with it so far :)

@q4a, sure. let me know

ugeen4 (dismissed):

src/apps/engine/src/compiler.cpp
```diff
@@ -1805,10 +1802,10 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
 
     if (bDebugInfo)
     {
-        strcpy_s(DebugSourceFileName, Segment.name);
-        fnsize = strlen(Segment.name);
+        strcpy_s(DebugSourceFileName, Segment.name.c_str());
+        fnsize = Segment.name.length();
         CompileToken(Segment, DEBUG_FILE_NAME, 3, (char *)&DebugSourceLine, sizeof(uint32_t), (char *)&fnsize,
-                     sizeof(uint32_t), Segment.name, strlen(Segment.name));
+                     sizeof(uint32_t), Segment.name, fnsize);
```
q4a: Please, use `Segment.name.c_str()` to avoid warning/error:
```
error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char>') through variadic method; call will abort at runtime [-Wnon-pod-varargs]
                     sizeof(uint32_t), Segment.name, fnsize);
                                               ^
```

q4a:
src/apps/engine/src/compiler.cpp - line 3736.
Please, replace
`SetError("Segment (%s) not loaded", SegmentTable[segment_index].name);`
with
`SetError("Segment (%s) not loaded", SegmentTable[segment_index].name.c_str());`
to avoid warning-error:
```
error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char>') through variadic method; call will abort at runtime [-Wnon-pod-varargs]
            SetError("Segment (%s) not loaded", SegmentTable[segment_index].name);
                                                                            ^
```

q4a (changes requested):

src/apps/engine/CMakeLists.txt
```diff
@@ -1,8 +1,13 @@
...
+
 STORM_SETUP(
     TARGET_NAME engine
     TYPE executable
-    DEPENDENCIES     
-	
+    DEPENDENCIES
+    
+    # this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
+    $<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>
```
q4a: Can you explain why you need to link mimalloc before c-runtime?
And if you really need it - try to avoid direct filenames: for example debug linux lib called `libmimalloc-debug.so`

q4a (changes requested):

src/apps/engine/CMakeLists.txt
espkk: The idea is to override both C++ and C std memory handling. forC++ there is a well-defined way to override all variations of new/delete operators. 
However, for lower-level C alloc stuff it's not that simple. On windows, the allocator can only be injected at runtime.
This creates a constraint for linker/loader order. mimalloc-redirect shall be loaded strictly before C runtime (for modern windows it is ucrtbase/ucrtbased).

Regarding hardcoded names. I have no idea how we can deduce the names. I tried a few techniques and they do not work. We are heavily constrained by Conan API here.
One idea I have in mind - on Linux I think it can be statically linked. Therefore we can probably configure it at conanfile level depending on the platform. And disable the above code on Linux. But this is not something I can do myself right now. So feel free to either try it yourself or disable this code for your fork for now

src/apps/engine/src/compiler.cpp
espkk: good catch

src/apps/engine/CMakeLists.txt
espkk: good catch

q4a: I tested few vars like `${CONAN_LIBS_MIMALLOC_RELEASE}` and `${CONAN_PKG_LIBS_MIMALLOC_RELEASE}`, but they don't fit. So I can offer simple solution, which works fine on Windows and Linux
```
$ git diff
diff --git a/src/apps/engine/CMakeLists.txt b/src/apps/engine/CMakeLists.txt
index 1008563e..0a94b106 100644
--- a/src/apps/engine/CMakeLists.txt
+++ b/src/apps/engine/CMakeLists.txt
@@ -1,12 +1,18 @@
 include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)

+if (WIN32)
+set(MIMALLOC_DEP "$<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>")
+else()
+set(MIMALLOC_DEP "$<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/libmimalloc.so,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/libmimalloc-debug.so>")
+endif()
+
 STORM_SETUP(
     TARGET_NAME engine
     TYPE executable
     DEPENDENCIES

     # this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
-    $<IF:$<CONFIG:Release>,${CONAN_LIB_DIRS_MIMALLOC_RELEASE}/mimalloc.lib,${CONAN_LIB_DIRS_MIMALLOC_DEBUG}/mimalloc-debug.lib>
+    ${MIMALLOC_DEP}

     # internal
     common
```

espkk: I hope we'll find a better way... but let it be for now :)

q4a (approved):

q4a:
Look good. Thanks.

ugeen4 (approved):

PR #325, opened 2022-01-18
[core] Don't forward SDL events until core initialized

espkk:
Doesn't look neat yet the core still needs a lot of refactoring


q4a:
Thanks. Tested on Windows and Linux branch. The problem was solved.

ugeen4 (approved):

PR #327, opened 2022-01-18
CI: add skip-duplicate-actions

q4a:
Check this: https://github.com/marketplace/actions/skip-duplicate-actions
Looks like that this is better solution for skipping duplicate CI.
It this PR is good, then we can close https://github.com/storm-devs/storm-engine/pull/326

Tested on my repository: https://github.com/q4a/storm-engine/pull/53
it has 2 quick `pre_job` for `push` and `PR`.
`pre_job` for `push` returned `should_skip = true` real build was skipped.
`pre_job` for `PR` returned `should_skip = false` real build did the job.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -13,7 +13,22 @@ on:
       - '*.md'
 
 jobs:
+  pre_job:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


espkk (approved):
well, let's try this first
thanks for the PR!

PR #328, opened 2022-01-18
Replace GetKeyState and GetAsyncKeyState with core functions

q4a:
This is few cherry-picks from my linux branch with replaced GetKeyState and GetAsyncKeyState with core functions.

### DIFF
--- a/src/libs/ball_splash/src/ball_splash.cpp
+++ b/src/libs/ball_splash/src/ball_splash.cpp
@@ -115,7 +115,7 @@ void BALLSPLASH::Realize(uint32_t _dTime)
 
     RDTSC_E(ticks);
     /*
-    if ((GetKeyState('Z') & 0x8000) != 0)
+    if ((core.Controls->GetKeyState('Z') & 0x8000) != 0)
     {
       renderer->Print(0, 150, "splash: all = %d, count = %d", ticks/1000, TSplash::processCount);
       if (TSplash::processCount)
--- a/src/libs/common/include/controls.h
+++ b/src/libs/common/include/controls.h
@@ -192,6 +192,16 @@ class CONTROLS
 
     virtual void SetMouseSensivityY(float){};
 
+    virtual short GetAsyncKeyState(int vk)
+    {
+        return GetAsyncKeyState(vk);
+    }
+
+    virtual short GetKeyState(int vk)
+    {
+        return GetKeyState(vk);
+    }
+
     virtual short GetDebugAsyncKeyState(int vk)
     {
         return GetAsyncKeyState(vk);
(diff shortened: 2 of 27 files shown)


espkk:
hey @q4a , could you clarify the status of this PR?

q4a:
@espkk Changes in this PR is wrong. Here is few messages about it from discord:
```
mitrokosta — 18.01.2022:
I agree with changing GetKeyState to core functions, but GetKeyState should not be replaced with GetDebugKeyState
The difference is in Debug part
And it is significant

q4a — 18.01.2022:
Ok, I did not know that. Should I create new function in core?

mitrokosta — 18.01.2022:
No, just wait a bit, I’ll explain
GetAsyncKeyState should be replaced with
core.Controls->IsKeyPressed
because core.Controls->GetDebugAsyncKeyState works only if ondebugkeys is enabled in engine.ini
otherwise it returns 0
and keep in mind IsKeyPressed returns bool, not -1/0
Actually you can create GetAsyncKeyState/GetKeyState so you have to do less manually
this GetAsyncKeyState should use IsKeyPressed internally
you should look at PCS_CONTROLS::GetDebugAsyncKeyState to know more

q4a — 18.01.2022:
Yes, I understand my problem - I always looked on virtual functions:
    virtual short GetDebugAsyncKeyState(int vk)
    {
        return GetAsyncKeyState(vk);
    }

    virtual short GetDebugKeyState(int vk)
    {
        return GetKeyState(vk);
    }
PCS_CONTROLS::GetDebugAsyncKeyState is used in real code

q4a — 18.01.2022:
There is few calls like if (LOWORD(GetKeyState(VK_NUMLOCK)) != 0) like in https://github.com/storm-devs/storm-engine/blob/develop/src/libs/location/src/location_camera.cpp#L565
It's not pressed, but toggled. So I need to think..
If the low-order bit is 1, the key is toggled. - https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeystate

mitrokosta — 18.01.2022:
Hm, we might need an adjustment in IsKeyPressed
We need to find out how numlock state is handled in SDL
we can probably try something like this (SDL_GetModState() & KMOD_NUM) != 0
but encapsulated
```

So problem is more complex and this changes is wrong. I will make better version, but later. I'm focused with render/graphics part now. If you don't like unmerged draft - then close this one, I'll create better one and make another PR.

espkk:
@q4a
I remember the issue, also I don't see any problem with having an unmerged draft until it's fully abandoned :)
I just wanted to clarify your steps regarding this PR. So, thanks for the explanation 👍 
It's up to you to proceed inside this or in a new one. Just let me know if you need any help with this or if you don't plan to finish it.

q4a:
@espkk @mitrokosta - I reworked the changes. Check it, please.

q4a:
I fixed VK_NUMLOCK checking and free_camera now works:
free_camera-fix

mitrokosta (approved):

espkk (approved):
cannot test this rn but looks good, thanks!

PR #329, opened 2022-01-18
Add script caching for faster loading

mitrokosta:
Turned off by default, to turn on add ``use_cache = 1`` in ``script`` section of engine.ini. Creates compiled files in PROGRAM\\_\_cached\_\_ directory.

### DIFF
--- a/src/apps/engine/src/compiler.cpp
+++ b/src/apps/engine/src/compiler.cpp
@@ -1420,6 +1420,10 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
                     SetError("define redefinition: %s", di.name);
                     return false;
                 }
+                if (use_script_cache_)
+                {
+                    script_cache_.defines.emplace_back(di.name, di.deftype, di.data4b);
+                }
             }
             break;
 
@@ -7471,7 +7475,7 @@ void COMPILER::SaveVariablesToCache(storm::script_cache::Writer &writer)
     }
 }
 
-void COMPILER::WriteFunctionsToCache(storm::script_cache::Writer &writer)
+void COMPILER::SaveFunctionsToCache(storm::script_cache::Writer &writer)
 {
     writer.WriteData(script_cache_.functions.size());
     for (auto &[fi, arguments, local_variables] : script_cache_.functions)
@@ -7501,7 +7505,7 @@ void COMPILER::WriteFunctionsToCache(storm::script_cache::Writer &writer)
     }
 }
 
-void COMPILER::WriteScriptLibrariesToCache(storm::script_cache::Writer &writer)
+void COMPILER::SaveScriptLibrariesToCache(storm::script_cache::Writer &writer)
 {
     writer.WriteData(script_cache_.script_libs.size());
     for (auto &lib_name : script_cache_.script_libs)
@@ -7510,7 +7514,7 @@ void COMPILER::WriteScriptLibrariesToCache(storm::script_cache::Writer &writer)
     }
 }
 
-void COMPILER::WriteEventHandlersToCache(storm::script_cache::Writer &writer)
+void COMPILER::SaveEventHandlersToCache(storm::script_cache::Writer &writer)
 {
     writer.WriteData(script_cache_.event_handlers.size());
     for (auto &[event, handler] : script_cache_.event_handlers)
@@ -7520,7 +7524,7 @@ void COMPILER::WriteEventHandlersToCache(storm::script_cache::Writer &writer)
     }
 }
 
-void COMPILER::WriteByteCodeToCache(storm::script_cache::Writer &writer, const SEGMENT_DESC &segment)
+void COMPILER::SaveByteCodeToCache(storm::script_cache::Writer &writer, const SEGMENT_DESC &segment)
 {
     auto code = std::string_view(segment.pCode, segment.pCode + segment.BCode_Program_size);
     writer.WriteBytes(code);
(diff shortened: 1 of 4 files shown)


ugeen4 (dismissed):
tested ! don't see problems at all

src/apps/engine/src/compiler.cpp
```diff
@@ -7,8 +7,11 @@
 
 #include "s_debug.h"
 #include "logging.hpp"
+#include "script_cache.h"
 #include "storm_assert.h"
 
+#include <map>
```
espkk: any reason why not `unordered_map`?

src/apps/engine/src/compiler.cpp
```diff
@@ -1583,6 +1632,10 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
                                             return false;
                                         }
                                         real_var->value->Set(1);
```
espkk: Can we extract kind of `SetVariable` function here accepting Varinfo and value, like this:
```
void Compiler::SetVariable(VarInfo * var, const auto& value) {
  var->value->Set(value);
  if(use_script_cache_) {
    script_cache_.variables.back().value->Set(value);
  }
]
```
in order to avoid much code bloating?

src/apps/engine/src/compiler.cpp
```diff
@@ -7117,6 +7273,509 @@ void COMPILER::PrintoutUsage()
...
+            writer.WriteBytes(reinterpret_cast<const char*>(value));
+            break;
+        }
+    }
+}
+
+bool COMPILER::LoadSegmentFromCache(SEGMENT_DESC &segment)
+{
+    const auto path = GetSegmentCachePath(segment);
+    if (!exists(path))
```
espkk: this may throw, please handle the exceptions

src/apps/engine/src/compiler.cpp
```diff
@@ -7117,6 +7273,509 @@ void COMPILER::PrintoutUsage()
...
+            break;
+
+        case FLOAT_NUMBER:
+            di.data4b = static_cast<uintptr_t>(*reader.ReadData<float>());
+            break;
+
+        case STRING: {
+            auto value = reader.ReadBytes();
+            const auto new_ptr = new char[value.size() + 1];
+            std::memcpy(new_ptr, value.data(), value.size());
```
espkk: prefer std::copy over std::memcpy

src/apps/engine/src/script_cache.h
```diff
@@ -0,0 +1,174 @@
...
+
+class Writer
+{
+public:
+    template <typename T>
+    void WriteData(const T &data)
+    {
+        const auto size = data_.size();
+        data_.resize(size + sizeof data);
+        std::memcpy(data_.data() + size, &data, sizeof(T));
```
espkk: prefer std::copy over std::memcpy

src/apps/engine/src/script_cache.h
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    void WriteBytes(std::string_view data)
+    {
+        WriteData(data.size());
+        const auto size = data_.size();
+        data_.resize(size + data.size());
+        std::memcpy(data_.data() + size, data.data(), data.size());
```
espkk: prefer std::copy over std::memcpy

src/apps/engine/src/compiler.cpp
```diff
@@ -1197,11 +1213,17 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
...
-        Segment.Files_list->AddUnicalString(file_name);
+        auto is_new = Segment.Files_list->AddUnicalString(file_name);
         file_code = Segment.Files_list->GetStringCode(file_name);
         pProgram = nullptr;
         Program_size = 0;
         pSegmentSource = LoadFile(file_name, SegmentSize);
+        if (is_new && use_script_cache_)
+        {
+            script_cache_.files.emplace_back(file_name);
+            script_cache_.crc = storm::script_cache::ComputeCRC(0, {});
```
espkk: why this is needed? it seems to make no difference

src/apps/engine/src/core_impl.cpp
```diff
@@ -252,6 +252,9 @@ void CoreImpl::ProcessEngineIniFile()
         Compiler->SetProgramDirectory(String);
     }
 
+    const auto use_cache = engine_ini->GetInt("script", "use_cache", false);
```
espkk: [script] section is read and processed straight from COMPILER class. it makes sense to move it there as well

espkk (changes requested):
Looks great, thanks! I can confirm this noticeably reduces loading time.
Since it's not enabled by default, I think we don't need to test it extensively before the merge
Just check the points below, please

src/apps/engine/src/compiler.cpp
mitrokosta: idk, fixed

src/apps/engine/src/compiler.cpp
mitrokosta: moved assignment in one spot

src/apps/engine/src/compiler.cpp
mitrokosta: done

src/apps/engine/src/compiler.cpp
mitrokosta: done with std::ranges::copy

src/apps/engine/src/compiler.cpp
mitrokosta: fixed

src/apps/engine/src/core_impl.cpp
mitrokosta: done

espkk:
Thanks, looks pretty good now! Let me test it again and I'll merge

espkk (approved):

PR #330, opened 2022-01-19
[core] fallback to 0 if from_chars failed

espkk:


mitrokosta (approved):
Thanks! Scripts relying on that behavior are working again!

PR #331, opened 2022-01-19
Replace few MSVC-only functions with std and fio

q4a:
This is few cherry-picks from my linux branch with:
- changing ptrdiff_t with std::ptrdiff_t.
- replacing few _* functions.
- replacing MSVC-only _ltoa with sprintf.
- add #include <optional>.
- restore few `unsigned long long`, because type should match with getDSPClock param from fmod.hpp.
- replacing `CreateFile`, `WriteFile`, `CloseHandle` with `fio` functions.
- removing unnecessary precreate directory, because `std::filesystem::create_directories` do it automatically.

### DIFF
--- a/src/apps/engine/src/compiler.cpp
+++ b/src/apps/engine/src/compiler.cpp
@@ -1845,12 +1845,11 @@ bool COMPILER::Compile(SEGMENT_DESC &Segment, char *pInternalCode, uint32_t pInt
         _splitpath(Segment.name, nullptr, nullptr, file_name, nullptr);
         strcat_s(file_name, ".b");
         std::wstring FileNameW = utf8::ConvertUtf8ToWide(file_name);
-        fh = CreateFile(FileNameW.c_str(), GENERIC_WRITE, FILE_SHARE_READ, nullptr, CREATE_ALWAYS,
-                        FILE_ATTRIBUTE_NORMAL, nullptr);
-        if (fh != INVALID_HANDLE_VALUE)
+        auto fileS = fio->_CreateFile(file_name, std::ios::binary | std::ios::out);
+        if (fileS.is_open())
         {
-            WriteFile(fh, Segment.pCode, Segment.BCode_Program_size, (LPDWORD)&dwR, nullptr);
-            CloseHandle(fh);
+            fio->_WriteFile(fileS, Segment.pCode, Segment.BCode_Program_size);
+            fio->_CloseFile(fileS);
         }
     }
     return true;
--- a/src/apps/engine/src/core_impl.cpp
+++ b/src/apps/engine/src/core_impl.cpp
@@ -238,7 +238,7 @@ bool CoreImpl::Initialize()
 
 void CoreImpl::ProcessEngineIniFile()
 {
-    char String[_MAX_PATH];
+    char String[MAX_PATH];
 
     bEngineIniProcessed = true;
 
--- a/src/libs/lighter/src/l_geometry.cpp
+++ b/src/libs/lighter/src/l_geometry.cpp
@@ -466,7 +466,7 @@ bool LGeometry::Save()
             if (dir[p] == '\\')
             {
                 dir[p] = 0;
-                if (_access(dir, 0) == -1)
+                if (!fio->_FileOrDirectoryExists(dir))
                 {
                     if (!fio->_CreateDirectory(dir))
                     {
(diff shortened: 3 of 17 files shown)


src/apps/engine/src/compiler.cpp
```diff
@@ -6375,11 +6374,11 @@ void COMPILER::SaveVariable(DATA *pV, bool bdim)
         break;
     case VAR_AREFERENCE:
         pString = nullptr;
-        __try
+        try
```
espkk: This won't work unless we specify `/EHa`. And we really don't want to do that

src/apps/engine/src/core_impl.cpp
```diff
@@ -238,7 +238,7 @@ bool CoreImpl::Initialize()
 
 void CoreImpl::ProcessEngineIniFile()
 {
-    char String[_MAX_PATH];
+    char String[MAX_PATH];
```
espkk, line 241: I'm not sure if MAX_PATH on Linux has the same meaning. Can we check that?

src/libs/renderer/src/screenshot.cpp
```diff
@@ -29,7 +29,7 @@ void DX9RENDER::SaveCaptureBuffers()
     for (fi = iCaptureFrameIndex; fi < iCaptureFrameIndex + 10000; fi++)
     {
         sprintf_s(cFileName, "k3cap_%04d.tga", fi);
-        if (_access(cFileName, 0) == -1)
+        if (access(cFileName, 0) == -1)
```
espkk: access is deprecated, will cause warnings or CTE depending on level
we can implement it in file service or try using `std::filesystem::permissions` instead

espkk (changes requested):

src/apps/engine/src/compiler.cpp
espkk: Also, I'm wondering how that *frankly speaking awful* check can be done with Linux. Are you aware?

mitrokosta: Please roll this back unless you come up with an idea on how to save arefs without relying on catching access violations. If this goes through without any changes we will have crashes on game save back.

q4a: done

src/libs/renderer/src/screenshot.cpp
q4a: `_access(cFileName, 0)` - checking if file/dir exists, so I replaced it with `std::filesystem::exists`

src/apps/engine/src/core_impl.cpp
q4a: I'm sure, that `MAX_PATH` == `_MAX_PATH` on Windows. So it will be easier to fix it for Linux in future.
In fact on Linux there is only `#define PATH_MAX 4096` in `limits.h`.

src/libs/renderer/src/screenshot.cpp
```diff
@@ -29,7 +29,7 @@ void DX9RENDER::SaveCaptureBuffers()
     for (fi = iCaptureFrameIndex; fi < iCaptureFrameIndex + 10000; fi++)
     {
         sprintf_s(cFileName, "k3cap_%04d.tga", fi);
-        if (_access(cFileName, 0) == -1)
+        if (!std::filesystem::exists(cFileName))
```
espkk: it may throw. please do exception check or use overload:
```
bool exists( const std::filesystem::path& p, std::error_code& ec )
```

src/libs/lighter/src/l_geometry.cpp
```diff
@@ -466,7 +466,7 @@ bool LGeometry::Save()
             if (dir[p] == '\\')
             {
                 dir[p] = 0;
-                if (_access(dir, 0) == -1)
+                if (!std::filesystem::exists(dir))
```
espkk: may throw

espkk (changes requested):
all's good now, except for std::filesystem::exists, please check it

src/libs/renderer/src/screenshot.cpp
q4a: Thanks. Done.

src/libs/lighter/src/l_geometry.cpp
q4a: Thanks. Done.

espkk (approved):
Great, thanks

PR #332, opened 2022-01-20
[core] fix mimalloc fun deinit order dependency

espkk:


ugeen4 (approved):
Tested! Now i can see non-empty mimalloc.log

PR #333, opened 2022-01-20
[xinterface] Correctly use font scaling factor when calculating string width

espkk:


ugeen4 (approved):
Tested, don't see any problems.Thanks!

PR #334, opened 2022-01-20
Replace D3DXQUATERNION and D3DXMATRIX

q4a:
Replace D3DXQUATERNION and D3DXMATRIX with internal functions from math3d/quaternion.h and math3d/matrix.h


espkk:
We need to measure performance before replacing quaternions
The difference potentially may be high
Could you do the measurements?

q4a:
> Could you do the measurements?

I tested msvc-release on my i5-3450 + GT 730 - start new game and tested on island, in open sea, in storm and in ship fight: get same 62-63 fps before and after this changes.
OIs there another tests or conditions, that I need to check?

ugeen4:
did you use vsink=on mode or not ?

espkk:
> > Could you do the measurements?
> 
> I tested msvc-release on my i5-3450 + GT 730 - start new game and tested on island, in open sea, in storm and in ship fight: get same 62-63 fps before and after this changes. OIs there another tests or conditions, that I need to check?

60 is very low for such tests... I would compare at least at 200
regarding Quaternion, I made some profiling and it seems to be not worse in terms of performance. that's good
So we only need to check sea, sky performance, and also make sure there is no difference in calculations

q4a:
> did you use vsink=on mode or not ?

Yes, sorry. I checked only max_fps. With `vsync = 0` I got same fps before and after patch:
95-115 - on island.
400-460 - sea map.
120-130 - in sea battle.

q4a:
Please, don't merge it for now - i need to do some more tests.

q4a:
Looks like it ready for merge

espkk (approved):
Thanks! I see no issue, but I would ask someone else to check as well since the code is not trivial

Hammie (approved):
Looks good!

q4a:
It's time to merge?)

PR #335, opened 2022-01-21
[core] fix memory leaks

olefirenque:
I checked some delete statements for correctness and it seems like there are many PTRs that own memory allocated with new[]. 
Hope it might help.


espkk (approved):
Thanks for the PR!
They would not cause memory leak, but it's always good to have proper methods used

PR #338, opened 2022-01-23
battle_ship_sign: fix for custom sized texture

Artess999:
Fix for custom sized texture (feature that I've added earlier), in some cases ship icons was incorrect on the sea


espkk (approved):

ugeen4 (approved):
good catch! thanks!

PR #342, opened 2022-02-09
Fix '#' prefix in scroll-image text

Hammie:
Small fix in the scroll-image code. This was causing issues while running PotC/NH, and with neither CoAS or TEHO using scroll-image I suspect it is a refactoring bug that sneaked past.


espkk (approved):
makes sense, thanks

PR #344, opened 2022-03-06
Message refactor (part 2)

Hammie:
Some more refactoring of the message system. Replaces the variable arguments with a variadic template function.

This makes it possible to pass the message directly as a parameter, makes it easier to emit an event without parameters, and makes it possible to pass `std::string` and `std::string_view` directly.

I also did some experiments with moving the format string to a literal template argument, as this would allow some very useful compile-time checks, but I felt the required C++ features were not quite well-supported enough yet, so these were excluded for now.

Tested with TEHO 1.7.3 and everything seemed to work as before.


np-c0mp1ete (approved):
No comments from looking at the diff, thanks for the MR!

espkk (approved):

PR #346, opened 2022-03-17
blender-plugins: fix for blender 3.1

Artess999:


np-c0mp1ete (approved):
Thanks for the update!

PR #351, opened 2022-04-01
[core] Fix script cache crash

espkk:


mitrokosta (approved):
doesn't break anything, thx

PR #352, opened 2022-04-08
[sea_ai] ai_group configurability

espkk:
- individual ship position in group can be manipulated through event
- removed non-vanilla behavior (in any case it was broken)

### DIFF
--- a/src/libs/sea_ai/src/ai_group.cpp
+++ b/src/libs/sea_ai/src/ai_group.cpp
@@ -2,8 +2,7 @@
 #include "shared/messages.h"
 
 std::vector<AIGroup *> AIGroup::AIGroups;
-float AIGroup::fDistanceBetweenGroupShips = 300.0f;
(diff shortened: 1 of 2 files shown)


espkk:
Script snippet for a default pos calculation:
```#define DISTANCE_BETWEEN_GROUP_SHIPS 250.0

#event_handler("Group_GetShipPosDefault", "Group_GetShipPosDefault");

ref Group_GetShipPosDefault()
{
  int n = GetEventData();
  float x = GetEventData();
	float y = GetEventData();
	float z = GetEventData();
    
  float dist = DISTANCE_BETWEEN_GROUP_SHIPS * n;

  float pos[3];
  
  pos[0] = x - dist * sin(y);
  pos[1] = y;
  pos[2] = z - dist * cos(y);
  
  return &pos;
}
```

Also group commander's attribute list should have the attribute `GroupShipPos_event` set to a desired event for a particular group. e.g. Group_GetShipPosDefault for a default formation

PR #354, opened 2022-04-14
[core] internal_functions: sort (attributes)

espkk:


Hammie (approved):
Looks good.

I do have some concern about case-insensitivity where it may lead to strange results, because I'm pretty sure attribute names are not case-normalized.

espkk:
Actually, they are case-normalized if used from scripts: https://github.com/storm-devs/storm-engine/blob/0a36c919d09cd3c903ff328341a5e1caad797ec8/src/apps/engine/src/compiler.cpp#L3424
So, I expect it to be fine in general...

PR #355, opened 2022-04-14
Modernize ATTRIBUTES class

Hammie:
I refactored `ATTRIBUTES` to try and clarify ownership a bit more and make things more const-correct. Tested with TEHO and everything still seems to work just fine.

I'm also planning to do a Part 2 for this, reworking the actual interface so it's a lot more pleasant to work with. But since those changes will be more opinionated I'd like to get this merged first.

### DIFF
--- a/src/apps/engine/src/core_impl.cpp
+++ b/src/apps/engine/src/core_impl.cpp
@@ -672,7 +672,7 @@ ATTRIBUTES *CoreImpl::Entity_GetAttributeClass(entid_t id_PTR, const char *name)
     return pE->AttributesPointer->FindAClass(pE->AttributesPointer, name);
 }
 
-char *CoreImpl::Entity_GetAttribute(entid_t id_PTR, const char *name)
+const char *CoreImpl::Entity_GetAttribute(entid_t id_PTR, const char *name)
 {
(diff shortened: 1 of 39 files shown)


espkk:
I looked through it several times and it looks good, but I'd be very careful with such a crucial part of the code. 
I had rewritten ATTRIBUTES code a few times but finally, I restored the original interface and part of the logic basically for two reasons, compatibility problems and performance. E.g. I remember a huge overhead of `std::optional`, and rewriting leafs as a linked list also did some performance penalty.

But it was years ago, `std::optional` was optimized I hope and the rest might be also good. I also would like to try SSO and/or some custom allocator here, despite we already have mimalloc as a general-purpose allocator.

Do you think we can prepare some UT and performance benchmarks for it?

PS
Oh, I used GitHub back-merge but it seems to be not enough since I used `friend` and the internals were changed. I'll rebase it a bit later

espkk (approved):
I did some cleanup and checked everything once again and did naive checks. I think it's fine to merge this. And I have an idea for UT/perf checks for this, but I need to prepare some structure refactoring before doing that.

If everything is ok with cleanup commit then feel free to merge.

Hammie:
@espkk Thanks for that!

PR #358, opened 2022-04-19
Fix crash due to string conversion from nullptr

Hammie:
Reverts some code to what it was previously


espkk (approved):
Though I'm trying to prepare some generic approach to prevent similar issues (this seems to be just one of them), this is good to keep the main branch stable. Thanks!

PR #361, opened 2022-04-27
Fix mimalloc_fun for Windows and Linux

q4a:
Old behavior.
In Windows and Linux `mimalloc_fun` will be called >= 2 times:
1) Initialization: will remove old file and will keep alive `mimalloc_log_path` until you exit.
2) Not sure, but I did not see in Windows any other messages, but it's totally possible and I have at least one in Linux.
3) When you press exit you: firstly go to the end of main `return EXIT_SUCCESS;` and after that calling `mimalloc_fun`

In Windows after exiting main and running `mimalloc_fun` I got `mimalloc_log_path.empty() == true`: it remove all previous messages and write only last block of messages.
In Linux after exiting main and running `mimalloc_fun` I got `mimalloc_log_path.empty() == false`, but it contain random non ascii chars => it will not remove previous messages in mimalloc.log but will create random non ascii filename in current folder.

```
q@i5-3450:~/git/sd-teho-public$ ls
''$'\036\224\017\311\361''`'$'\357\235\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\005'
 engine-1_d3d9.log
 engine-1.dxvk-cache
 engine.ini
 LICENSE
''$'\274\265''N'$'\t\030''F'$'\343''v'$'\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\337\005'
 PROGRAM
 README.md
 resource
 SAVE
 shaders
''$'\340'''$'\f\213\004'
q@i5-3450:~/git/sd-teho-public$
```

New behavior.
1. For Windows and Linux `mimalloc_fun` will get "init" arg only once in main when calling `mi_register_output` and then it will remove old  mimalloc.log.
2. All other calls will write to mimalloc.log without removing it.
3. I have to remove `static` in `GetStashPath()` and from `static std::filesystem::path mimalloc_log_path;` to not get random non ascii in Linux.

### DIFF
--- a/src/apps/engine/src/main.cpp
+++ b/src/apps/engine/src/main.cpp
@@ -127,7 +127,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
     }
     mi_register_output(mimalloc_fun, nullptr);
     mi_option_set(mi_option_show_errors, 1);
-    mi_option_set(mi_option_show_stats, 1);
+    mi_option_set(mi_option_show_stats, 0);
     mi_option_set(mi_option_eager_commit, 1);
     mi_option_set(mi_option_eager_region_commit, 1);
     mi_option_set(mi_option_large_os_pages, 1);


src/apps/engine/src/main.cpp
```diff
@@ -125,7 +125,10 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR szCmdLine,
         MessageBoxA(nullptr, "Another instance is already running!", "Error", MB_ICONERROR);
         return EXIT_SUCCESS;
     }
-    mi_register_output(mimalloc_fun, nullptr);
+    std::string *mi_arg = new std::string("init");
```
espkk: I don't fully understand the meaning of mi_arg (especially destroy_at for it) 🤔 
if the intention is to remove file at first write, just put a `static bool` and check for it. unlike`std::filesystem::path` it won't be destroyed (because there is nothing to destroy...)

src/libs/util/include/storm/fs.h
```diff
@@ -11,7 +11,7 @@ using namespace std::filesystem;
 
 inline path GetStashPath()
 {
-    static path path;
+    path path;
```
espkk: `SHGetKnownFolderPath`/`CoTaskMemFree` are relatively heavy
instead, we can e.g. add a default boolean arg which will turn it to work with a stack variable if set.
OR to keep the path in a static buffer (but size is a problem in that case, could be up to 2^16 for Unicode)
OR even to inline the code manually into `mimalloc_fun` because we need such only there so far
Anyway, at some point I would like to rework all that crappy `fs.h` stuff

espkk (changes requested):

q4a:
@espkk I'm not sure if it's a good idea to record mimalloc stats after exiting WinMain. Do we really need mimalloc stats? I suggest turning it off rather than trying to fix it. If you still want it, then I'll leave it enabled only for Windows like:
```
#ifdef _WIN32
    mi_option_set(mi_option_show_stats, 1);
#else
    mi_option_set(mi_option_show_stats, 0);
#endif
```

espkk (approved):
let's disable it for now 😢 
thanks!

PR #362, opened 2022-04-30
Fix bug in CopyAttributes

Hammie:
Fixes a bug in CopyAttributes caused by the ATTRIBUTES refactor


espkk (approved):
Tested, looks good, thanks!

PR #363, opened 2022-04-30
[location] supervisor: push models in same point

espkk:


espkk:
unknown
1000 npc spawned in just a few locators

Hammie (approved):

PR #365, opened 2022-05-01
Remove duplicates from target link list

Asapin:


espkk (changes requested):
it makes sense to remove duplicates before `list(APPEND _SETUP_DEPENDENCIES "${global_modules_list}")`
and `list(REMOVE_DUPLICATES _SETUP_DEPENDENCIES)` should be removed in this case

espkk (approved):
Now looks good. Thanks for the PR!

PR #366, opened 2022-05-09
Linux related changes part 1

q4a:
The Linux version is ready and works well enough in my opinion, so it's time to merge.
I decided to split the changes into several parts. In this (first) part there will be small changes and additional functions that are needed for building and correct working engine on Linux.


Hammie (commented):
I think this looks good, but I'm not sure why you added trivial constructors to several structs?

Is that needed because of compiler issues?

q4a:
@Hammie Yes. Clang-11 requires them. For example, when I remove constructor `RS_LINE(){}` I got this error:
```
/home/q/git/storm-engine/src/libs/island/src/foam.cpp:138:25: error: no matching constructor for initialization of 'RS_LINE'
                RS_LINE rl, r2;
                        ^
/home/q/git/storm-engine/src/libs/common/include/dx9render.h:45:8: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
struct RS_LINE
       ^
```
or if I remove both RS_LINE constructors - I got:
```
In file included from /home/q/git/storm-engine/src/libs/sea_ai/src/ai_ship_cannon_controller.cpp:1:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/thread:40:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/stop_token:34:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/atomic:41:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/atomic_base.h:41:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/atomic_wait.h:49:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/std_mutex.h:39:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/system_error:41:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/stdexcept:39:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/string:55:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/basic_string.h:40:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/ext/alloc_traits.h:34:
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/alloc_traits.h:514:4: error: no matching function for call to 'construct_at'
          std::construct_at(__p, std::forward<_Args>(__args)...);
          ^~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/vector.tcc:115:21: note: in instantiation of function template specialization 'std::allocator_traits<std::allocator<RS_LINE>>::construct<RS_LINE, CVECTOR, const unsigned int &>' requested here
            _Alloc_traits::construct(this->_M_impl, this->_M_impl._M_finish,
                           ^
/home/q/git/storm-engine/src/libs/sea_ai/src/ai_ship_cannon_controller.cpp:611:23: note: in instantiation of function template specialization 'std::vector<RS_LINE, std::allocator<RS_LINE>>::emplace_back<CVECTOR, const unsigned int &>' requested here
                Lines.emplace_back(vPos + 5.0f * cannon.GetDir(), red);
                      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_construct.h:94:5: note: candidate template ignored: substitution failure [with _Tp = RS_LINE, _Args = <CVECTOR, const unsigned int &>]: no matching constructor for initialization of 'RS_LINE'
    construct_at(_Tp* __location, _Args&&... __args)
```

Hammie (dismissed):
@q4a Interesting, I think these constructors are not the best way to solve the issue, but I don't actually have time to investigate this for a few more days.

For now I can say that `std::move(int, float, bool, etc..)` is generally discouraged, because it does not differ from a normal copy/assignment. But then again it also does not really matter.

I'd be happy to merge this as-is, and investigate the constructor issues later, but I'd also like an additional pair of eyes to have a quick look at this MR.

q4a:
@Hammie I'm not sure about `std::move` too, but I saw this commit im storm repo:
https://github.com/storm-devs/storm-engine/commit/16cfeca5e32f984870732b4a09f7939b839cc51b#diff-56897843cd404b3211bfc8ddb9e91c040517394471ec97a00e52320936058c89R71
```
    SLIBHOLDER(SCRIPT_LIBRIARY *library, std::string name)
        : library(library), name(std::move(name))
    {
    }
```
But that `std::move` was for `std::string`.
Anyway. I made separate commit for `std:move` and I can revert/squash it with any suitable fix: https://github.com/storm-devs/storm-engine/pull/366/commits/44ea6e50ccbba67c9b4c8b608d038aa5e3ae1859

`I'd be happy to merge this as-is, and investigate the constructor issues later, but I'd also like an additional pair of eyes to have a quick look at this MR.`
- yes, me too. My main goal is add Linux support without breaking anything on Windows - so I'm not hurry, I'm here and ready to improve or split this changes.

espkk:
@q4a thanks for the PR

I'm extremely busy these days, so I cannot review this properly (that's also the reason why I'm inactive in discord lately) :(
Please hold this back a bit

PS
to the point of discussion:
1) applying std::move to primitive types should be avoided (yea for string/vector/etc. nay for bool/float/etc.)
2) I think clang requires those constructors because it doesn't support that c++20 feature until clang-13. I would suggest not to introduce constructors for aggregate types. instead, try using brace (aggregate) initialization
e.g. instead of `emplace(x,y)` try `emplace({x,y})`

q4a:
@espkk @Hammie Ok, I know, that trivial constructors for structs is not so good, but I did not know better ways. I'll check if Clang-14 can handle it or will try use `emplace({x,y})` as espkk supposed, but I'll include it in next Linux PR.

Related commits was removed now.

src/libs/common/include/v_file_service.h
```diff
@@ -7,6 +7,16 @@
 #include <string>
 #include <vector>
 
+static bool starts_with(const std::string &str, const std::string &prefix)
```
espkk: `static`->`inline`?

src/libs/common/include/v_file_service.h
```diff
@@ -7,6 +7,16 @@
...
+{
+    return str.size() >= prefix.size() && 0 == str.compare(0, prefix.size(), prefix);
+}
+
+static bool ends_with(const std::string &str, const std::string &suffix)
```
espkk: `static`->`inline`?

src/libs/core/src/compiler.cpp
```diff
@@ -822,7 +822,7 @@ bool COMPILER::BC_LoadSegment(const char *file_name)
     // const auto len = strlen(file_name) + 1;
     // SegmentTable[index].name = new char[len];
     // memcpy(SegmentTable[index].name, file_name, len);
-    SegmentTable[index].name = _strdup(file_name);
+    SegmentTable[index].name = strdup(file_name);
```
espkk, line 825: this is POSIX and therefore `strdup` deprecated in MSVC

src/libs/island/src/foam.cpp
```diff
@@ -53,7 +53,7 @@ bool CoastFoam::Init()
 
 void CoastFoam::Execute(uint32_t Delta_Time)
 {
-    bEditMode = (bCanEdit) ? LOWORD(core.Controls->GetDebugKeyState(VK_NUMLOCK)) != 0 : false;
+    bEditMode = (bCanEdit) ? (core.Controls->GetDebugKeyState(VK_NUMLOCK) & 1) : false;
```
espkk: Original meaning not preserved here: it checked for non-zero in low word while your code checks only rightmost bit

src/libs/location/src/location_camera.cpp
```diff
@@ -563,7 +563,7 @@ void LocationCamera::ExecuteTopos(float dltTime)
 // Free flying camera execution
 void LocationCamera::ExecuteFree(float dltTime)
 {
-    if (LOWORD(GetKeyState(VK_NUMLOCK)) != 0)
+    if (GetKeyState(VK_NUMLOCK) & 1)
```
espkk: Original meaning not preserved here: it checked for non-zero in low word while your code checks only rightmost bit

src/libs/common/include/defines.h
```diff
@@ -104,6 +105,24 @@ constexpr float PId4 = (PI / 4.0f);
 #undef SQR
 #endif
 
+inline void toupr(char *str)
```
espkk: this could return `str` to comply with `_strupr`...

src/libs/sea_cameras/src/free_camera.cpp
```diff
@@ -90,7 +90,7 @@ void FREE_CAMERA::Move(uint32_t DeltaTime)
 {
     if (!isActive())
         return;
-    if (LOWORD(GetKeyState(VK_NUMLOCK)) != 0)
+    if (GetKeyState(VK_NUMLOCK) & 1)
```
espkk: same here

src/libs/renderer/src/s_device.cpp
```diff
@@ -1435,13 +1435,9 @@ bool DX9RENDER::TextureLoad(int32_t t)
...
     }
 
-    std::string_view name_sv(Textures[t].name, strlen(Textures[t].name));
-    auto has_resource_prefix =
-        std::distance(std::ranges::search(name_sv, "resource\\textures\\"sv, storm::detail::is_iequal{}).begin(),
-                      name_sv.begin()) == 0;
-    auto has_tx_postfix =
-        std::distance(std::ranges::search(name_sv, ".tx"sv, storm::detail::is_iequal{}).begin(),
-                      name_sv.end()) == 3;
+    tolwr(Textures[t].name);
```
espkk: if we do this how would we use upper register texture name on case-sensetive filesystem?

src/libs/core/src/compiler.cpp
```diff
@@ -6563,8 +6564,9 @@ bool COMPILER::SaveState(std::fstream &fileS)
     {
         char *pDst = new char[dwCurPointer * 2];
         uint32_t dwPackLen = dwCurPointer * 2;
+        uLongf tmpLong = dwPackLen;
         RDTSC_B(dw2);
-        compress2((Bytef *)pDst, (uLongf *)&dwPackLen, (Bytef *)pBuffer, dwCurPointer, Z_BEST_COMPRESSION);
+        compress2((Bytef *)pDst, &tmpLong, (Bytef *)pBuffer, dwCurPointer, Z_BEST_COMPRESSION);
```
espkk: the output is written to tmpLong but dwPackLen is used afterwards.
this should write to dwPackLen instead

espkk (changes requested):

src/libs/core/src/compiler.cpp
q4a: Is that problem? I can add `#define _strdup strdup` in Linux header - that will be good ?

src/libs/renderer/src/s_device.cpp
q4a: My `starts_with`/`ends_with` is case sensitive. So I just make `Textures[t].name` in low case to make sure that it will find "resource\\textures\\" or ".tx".
Reading in case sensetive filesystem will be in next PRs.

src/libs/island/src/foam.cpp
q4a: My mistake. I will fix it in this PR: https://github.com/storm-devs/storm-engine/pull/328

src/libs/location/src/location_camera.cpp
q4a: My mistake. I will fix it in this PR: https://github.com/storm-devs/storm-engine/pull/328

src/libs/sea_cameras/src/free_camera.cpp
q4a: My mistake. I will fix it in this PR: https://github.com/storm-devs/storm-engine/pull/328

src/libs/common/include/v_file_service.h
q4a: Done.

src/libs/common/include/v_file_service.h
q4a: Done.

src/libs/common/include/defines.h
q4a: Done.

src/libs/core/src/compiler.cpp
q4a: Fixed.

q4a:
@espkk It's ready for new review.

src/libs/renderer/src/s_device.cpp
espkk: the question is can it affect the system? originally, `Textures[t].name` contained a case-sensitive path
e.g. if the texture is being repeatedly requested but the matching search is case-sensitive, this would introduce superfluous resource allocation

espkk (changes requested):

src/libs/renderer/src/s_device.cpp
q4a: Changed. I hope this fix is good.

Hammie (approved):

espkk (approved):
Now looks good, thanks!

PR #371, opened 2022-05-22
Add basic Linux support: building, CI, not launching yet

q4a:
This PR still has some FIX_LINUX, but they are not critical:
- for intro video.
- for screenshots by pressing F8. Most important - I'll fix it ASAP.
- for crashpad.
- for Beeper and etc.

It still has no case sensitive file read/write support. I'll add it and some more stuff in next PR.
Most important in this PR - is full Linux building and CI for Linux. So I/we can check if new changes will break Linux.


.github/workflows/ci_linux.yml
```diff
@@ -0,0 +1,46 @@
+name: CI Linux
```
espkk, line 1: why cannot we this as a new job into cibuild.yml?
otherwise we'll need to duplicate all the common settings and also duplicate check job...

CMakeLists.txt
```diff
@@ -36,6 +36,30 @@ conan_cmake_run(CONANFILE conanfile.py
         steam=${STORM_ENABLE_STEAM}
 )
 
+if (NOT WIN32)
```
espkk, line 39: Can we put it into our conan artifactory instead?

cmake/StormSetup.cmake
```diff
@@ -224,7 +224,9 @@ macro(STORM_SETUP)
     foreach(module ${global_modules_list})
       if(TARGET ${module})
         # TODO: make it portable? (at least for gcc it is --whole-archive)
-        list(APPEND target_link_flags "/WHOLEARCHIVE:${module}")
+        if (WIN32)
```
espkk, line 230: does it work without `-force_load` for clang?

conanfile.py
```diff
@@ -16,27 +16,38 @@ class StormEngine(ConanFile):
...
 
     # optional dependencies
     def requirements(self):
+        if self.settings.os == "Windows":
+            # conan-center
+            self.requires("7zip/19.00")
+        else:
+            # conan-center
+            self.requires("libsafec/3.6.0")
+            self.requires("openssl/1.1.1o")
```
espkk: why we need openssl? 😮

conanfile.py
```diff
@@ -16,27 +16,38 @@ class StormEngine(ConanFile):
...
     default_options = {
         "sdl2:sdl2main": False,
+        "sdl2:pulse": False,
         "sentry-native:backend": "crashpad",
-        "sentry-native:transport": "winhttp",
         "mimalloc:shared": True,
         "mimalloc:override": True
     }
 
+    def configure(self):
```
espkk: this is redundant; transport should be configured properly with its default settings, so it's better to remove the whole function

extern/portable-snippets/COPYING.md
```diff
@@ -0,0 +1,124 @@
+Each source file contains a preamble explaining the license situation
```
espkk: I don't see any requirement to have this license file in the sources... License itself is already mentioned in the header file

extern/portable-snippets/debug-trap.h
```diff
@@ -0,0 +1,83 @@
+/* Debugging assertions and traps
```
espkk: this probably should be placed inside `libs/common` because it's already managed by our cmake system

extern/wine/COPYING.LIB
```diff
@@ -0,0 +1,502 @@
+		  GNU LESSER GENERAL PUBLIC LICENSE
```
espkk: same thing here - I don't see such requirement. we don't want unnecessary bloat
If we really need to supply licenses, I'd suggest to create a separate file mentioning all the external sources we use directly with links to corresponding licenses

extern/wine/winuser.rh
```diff
@@ -0,0 +1,1401 @@
+/*
```
espkk: this can be placed into `libs/common` as well if it's used globally

src/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ if (MSVC)
     add_link_options("$<$<CONFIG:Release>:/OPT:ICF>") # Perform identical COMDAT folding, needs to be explicitly specified with /DEBUG
     add_link_options("$<$<CONFIG:Release>:/LTCG>") # Enable link-time code generation
 else()
+    add_definitions(-D_GLIBCXX_USE_TBB_PAR_BACKEND=0) # Don't use tbb
```
espkk: could you please explain why this is needed? (would be great if you put the explanation directly into CMakeLists.txt)

src/apps/engine/src/main.cpp
```diff
@@ -119,8 +119,12 @@ void HandleWindowEvent(const storm::OSWindow::Event &event)
 int main(int argc, char *argv[])
 {
     // Prevent multiple instances
+#ifdef _WIN32 // CreateEventA
     if (!CreateEventA(nullptr, false, false, "Global\\FBBD2286-A9F1-4303-B60C-743C3D7AA7BE") ||
         GetLastError() == ERROR_ALREADY_EXISTS)
+#else
+    if (false)
```
espkk: the whole if-block can be surrounded with `ifdef/endif` without `else`, would be clearer

src/libs/common/include/storm_d3dx9.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#ifndef _WIN32
+
+#include <d3d9.h>
+#include <algorithm>
```
espkk: why `<algorithm>` needed here?

src/libs/common/include/storm_platform.h
```diff
@@ -0,0 +1,88 @@
...
+#include "storm_d3dx9.h"
+#include "wine/winuser.rh"
+#include "safe_str_lib.h"
+#undef EXTERN // fix for token.h:72:5: error: expected identifier EXTERN,
+
+#define MAX_PATH PATH_MAX
+#define _MAX_FNAME NAME_MAX
+#define MAKELONG(low, high) ((int32_t)(((uint16_t)(low)) | (((uint32_t)((uint16_t)(high))) << 16)))
+
+inline char *convert_path_sep(const char *cPath)
```
espkk: don't make those inline. instead, put the implementation into `common/src`. in that way you won't need to bloat compilation with headers and doubtful defines, moving them into cpp file

`#undef` (especially multiple) is also not a good idea

src/libs/core/src/compiler.cpp
```diff
@@ -6454,7 +6482,11 @@ void COMPILER::SaveVariable(DATA *pV, bool bdim)
         {
             pA = TraceARoot(pV->AttributesClass, pString);
         }
+#ifdef _WIN32 // __except
```
espkk: put some TODO tag here like FIX_LINUX
this needs to be fixed eventually because it'll lead to crashes otherwise...

src/libs/renderer/src/s_device.h
```diff
@@ -460,7 +466,11 @@ class DX9RENDER : public VDX9RENDER
 
     float FovMultiplier{ 1.0f };
 
+#ifdef _WIN32 // Effects
     Effects effects_;
+#else
+    CTechnique *pTechnique;
```
espkk: please use std::unique_ptr

espkk (changes requested):

.github/workflows/ci_linux.yml
q4a: It's possible. I think that there will be very few common settings, and when divided into different files, it will be possible:
1) fix CI for each OS separately without requiring a rebuild for another OS.
2) quickly understand which build fails just by looking at the `README.md`

Do you still think, that it should be into cibuild.yml?

CMakeLists.txt
q4a: Yes it is possible. I don't know conan very well and I think that it overcomplicated. If you can do it - I'm ready to add it, if not - then I would leave this part as it is, because conan simplifies building under Windows, but not under Linux.

cmake/StormSetup.cmake
q4a: As I wrote in the description - the main goal is full Linux building + CI. This is minimal fix to get Linux linking, but proper fix will require 1 more change:
https://github.com/q4a/storm-engine/commit/8c6b63b55b5475fb5c3de9f6da5f2353f322c046
Should I include it in this PR or the next one?

conanfile.py
q4a: Because of stupid conan. It will require 2 different versions of openssl for building libcurl and pulseaudio: https://github.com/q4a/storm-engine/runs/6645482117?check_suite_focus=true
```
ERROR: Conflict in libcurl/7.78.0:
    'libcurl/7.78.0' requires 'openssl/1.1.1o' while 'pulseaudio/14.2' requires 'openssl/1.1.1n'.
    To fix this conflict you need to override the package 'openssl' in your root package.
CMake Error at cmake/conan.cmake:524 (message):
  Conan install failed='1'
```
I already added `"sdl2:pulse": False,` and I still don't understand: why it's going to build `pulseaudio`

cmake/StormSetup.cmake
q4a: I added https://github.com/q4a/storm-engine/commit/8c6b63b55b5475fb5c3de9f6da5f2353f322c046 to this PR.

conanfile.py
q4a: Removed.

src/apps/engine/src/main.cpp
q4a: Done.

extern/portable-snippets/COPYING.md
q4a: Removed.

extern/portable-snippets/debug-trap.h
q4a: Done.

extern/wine/COPYING.LIB
q4a: Removed.

extern/wine/winuser.rh
q4a: Done.

src/CMakeLists.txt
q4a: Removed. It was left by mistake after early experiments.

src/libs/common/include/storm_d3dx9.h
q4a: Moved to storm_d3dx9.cpp, where it actually used.

src/libs/common/include/storm_platform.h
q4a: Done. I hope that I fixed it correctly.

src/libs/core/src/compiler.cpp
q4a: Done.

src/libs/renderer/src/s_device.h
q4a: Done. I hope that I fixed it correctly.

q4a:
@espkk Hi. Please, check my last fixes.

.github/workflows/ci_linux.yml
q4a: I tested and added duplicate check job into Linux CI.

espkk (approved):
overall looks good to me!

Hammie (approved):

PR #373, opened 2022-05-27
Rework entity_manager (again)

espkk:
Entity manager reworked in order to:
- further improve performance
- remove static constraints for entity amount
- fix some rare crash issues
- make code more readable and maintainable

TODO:
`GetEntityIds` by layer type still does a copy of a lot of data in each frame. Perhaps this can be optimized.
`GetEntityIds` by layer type includes duplicates. This may potentially cause issues and unexpected behaviour.


Hammie (approved):
Thanks, looks good to me!

Nice to see entity management is no longer global

ugeen4 (approved):
Ok for me

PR #375, opened 2022-06-05
PCS_CONTROLS: allow to use numpad arrows if regular arrows are in options

q4a:
This will allow to use numpad arrows if regular arrows are specified in the options like here:
unknown

### DIFF
--- a/src/libs/pcs_controls/src/pcs_controls.cpp
+++ b/src/libs/pcs_controls/src/pcs_controls.cpp
@@ -639,6 +639,26 @@ bool PCS_CONTROLS::IsKeyPressed(int vk)
         pressed = input_->KeyboardSDLKeyState(SDL_SCANCODE_RETURN) ||
                   input_->KeyboardSDLKeyState(SDL_SCANCODE_KP_ENTER);
     }
+    else if (vk == VK_UP)
+    {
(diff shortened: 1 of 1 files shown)


Hammie:
Possible include a numlock check `(SDL_GetModState() & KMOD_NUM) == (KMOD_NUM)` ?

q4a:
> Possible include a numlock check `(SDL_GetModState() & KMOD_NUM) == (KMOD_NUM)` ?

Yes, I was thinking about it, but @mitrokosta wanted to check old behavior. So I'll update PR after his tests.

mitrokosta:
> > Possible include a numlock check `(SDL_GetModState() & KMOD_NUM) == (KMOD_NUM)` ?
> 
> Yes, I was thinking about it, but @mitrokosta wanted to check old behavior. So I'll update PR after his tests.

I checked. and it mattered. With NumLock OFF keypad keys should work as arrows. With NumLock ON keypad keys should work as digits.

q4a:
> With NumLock OFF keypad keys should work as arrows. With NumLock ON keypad keys should work as digits.

@mitrokosta I updated PR. It should work as you described.

Hammie (approved):

espkk (approved):

PR #376, opened 2022-06-07
sailors: add customizable sailors models

Artess999:
Add customizable sailors models
Basic usage example in scripts:
```
string sailorsModels[2] = {"Lowcharacters\\Lo_Man_Kamzol_2", "Lowcharacters\\Lo_Man_Kamzol_3"};

void Ship_Walk_Create()
{
	int charIndex = GetEventData();
	aref ship = GetEventData();

	ref ch = GetCharacter(charIndex);

	if (!CheckAttribute(ch, "ship.type"))
	{
		trace("Ship_Walk_Create err: " + ch.id + " have invalid ship");
		return;
	}

	if(ch.id == "0" || sti(ch.ship.type) == SHIP_NOTUSED)
	{
		trace("Ship_Walk_Create err: " + ch.id + " have invalid ship");
		return;
	}

	int Index= sti(ch.ship.type);
	int ri= sti(RealShips[Index].basetype);

	SendMessage(&Sailors, "lise", AI_MESSAGE_ADD_SHIP, ship, /*ShipsTypes[ri].name*/RealShips[Index].name, &sailorsModels);
}
```

### DIFF
--- a/src/libs/sailors/src/sailors.cpp
+++ b/src/libs/sailors/src/sailors.cpp
@@ -722,27 +722,8 @@ void ShipWalk::CreateNewMan(SailorsPoints &sailorsPoints)
 
     shipMan[current].modelID = core.CreateEntity("MODELR");
 
-    switch (rand() % 6) // 6 different character types
-    {
-    case 0:
-        core.Send_Message(shipMan[current].modelID, "ls", MSG_MODEL_LOAD_GEO, "Lowcharacters\\Lo_Man_1");
-        break;
-    case 1:
-        core.Send_Message(shipMan[current].modelID, "ls", MSG_MODEL_LOAD_GEO, "Lowcharacters\\Lo_Man_2");
-        break;
-    case 2:
-        core.Send_Message(shipMan[current].modelID, "ls", MSG_MODEL_LOAD_GEO, "Lowcharacters\\Lo_Man_3");
-        break;
-    case 3:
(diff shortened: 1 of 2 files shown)


espkk (dismissed):

ugeen4 (dismissed):
looks very interesting, thank you!

src/libs/sailors/src/sailors.cpp
```diff
@@ -1170,9 +1156,26 @@ uint64_t Sailors::ProcessMessage(MESSAGE &message)
...
+                {
+                    pvd->Get(pstr, i);
+                    shipManModels.push_back(pstr);
+                }
+            }
+        }
 
         shipWalk.emplace_back();
-        if (shipWalk[shipsCount].Init(shipID, editorMode, c.c_str()))
+        if (shipWalk[shipsCount].Init(shipID, editorMode, c.c_str(), shipManModels))
```
mitrokosta: I'd accept shipManModels by value, and move it into Init, rather than passing a reference and killing it from Init

espkk: copy-elision is not guaranteed here (and it doesn't actually happen). and having a method that accepts only rvalue-ref is as questionable as modifying object by lvalue-ref
but I'd go for it(rvalue-ref) anyway

espkk (approved):

PR #378, opened 2022-06-08
Sailors refactoring

espkk:
- some refactoring
 - paths should be more randomized now
 - should fix rare ShipMan crash


Hammie (approved):
Looks good to me!

I think it might make sense to introduce some kind of `frand()` function for easily getting random numbers in the [0.0-1.0] range. We seem to need that in quite a few places.

PR #382, opened 2022-07-10
Fix XI_FindFolders in Linux

q4a:
This fixes reading user/save profiles in Linux.


ugeen4 (approved):
checked, all ok! Thanks for PR !

espkk (approved):

PR #383, opened 2022-07-18
Fix crash when no lines need to be drawn in the battle navigator

Hammie:
Just a quick fix for a crash I encountered while debugging New Horizons

### DIFF
--- a/src/libs/battle_interface/src/utils.cpp
+++ b/src/libs/battle_interface/src/utils.cpp
@@ -436,10 +436,7 @@ void BILinesInfo::Init(VDX9RENDER *rs, ATTRIBUTES *pA)
 
 void BILinesInfo::Draw()
 {
-    if (!lines.empty())
-    {
-        pRS->DrawLines2D(&lines[0], lines.size() / 2, "Line");
-    }
+    pRS->DrawLines2D(std::data(lines), lines.size() / 2, "Line");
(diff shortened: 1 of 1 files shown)


q4a:
Looks good. Thanks!

espkk (commented):
Hey. I'm wondering how this caused a crash for you. `DrawLines2D` returns immediately if the second parameter==0
Therefore, I don't see any difference between `empty()` and `size() / 2`
But maybe I'm missing something...

Hammie:
@espkk It's a debug assertion triggered by doing `&lines[0]`, so probably only crashes in debug builds.

espkk:
> @espkk It's a debug assertion triggered by doing `&lines[0]`, so probably only crashes in debug builds.

Ah...you're right, it makes sense
Though we don't dereference memory the access operator cannot know that
It's very likely that operator[] is not allowed to be used on an empty array

However, we can safely put `std::data(lines)` instead, right? Or even pass a cref to a vector, but then we'll likely have to rewrite some other portions of code

I mean I also don't mind `empty()` check. But to me it just sounds consistent enough that we pass a buffer with zero objects to draw and it works fine

Hammie:
@espkk `data()` does work so I updated the code

espkk (approved):

PR #385, opened 2022-07-21
Add new LegacyDialog entity

Hammie:
Adds a LegacyDialog entity that implements the old Seadogs 2 dialog system for use in New Horizons.

Does not yet include pagination or mood settings for the head animation. 

Also includes https://github.com/storm-devs/storm-engine/pull/383


src/libs/dialog/src/legacy_dialog.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+        CMatrix mtx, view, prj;
+        uint32_t lightingState, zenableState;
+        RenderService->GetTransform(D3DTS_VIEW, view);
+        RenderService->GetTransform(D3DTS_PROJECTION, prj);
+        RenderService->GetRenderState(D3DRS_LIGHTING, &lightingState);
+        RenderService->GetRenderState(D3DRS_ZENABLE, &zenableState);
+
+        mtx.BuildViewMatrix(CVECTOR(0.0f, 0.0f, 0.0f), CVECTOR(0.0f, 0.0f, 1.0f), CVECTOR(0.0f, 1.0f, 0.0f));
+        RenderService->SetTransform(D3DTS_VIEW, (D3DXMATRIX *)&mtx);
```
q4a: Please, use `(D3DMATRIX *)` instead of `(D3DXMATRIX *)`

Hammie: Updated it

q4a:
@Hammie sorry, but clang is more strict. If you create `td::vector<LinkEntry> formattedLinks_;` then you need to emplace_back only LinkEntry. I fixed problem, builded on Linux and tested a bit.
Works fine, thanks!
https://github.com/storm-devs/storm-engine/pull/385/commits/2ff2085dfa50b6dbc245504482abb16bb938d3d1
```
             for (const auto &text : link_texts)
             {
-                formattedLinks_.emplace_back(text, static_cast<int32_t>(i));
+                formattedLinks_.emplace_back(LinkEntry{text, static_cast<int32_t>(i)});
             }
```

Hammie:
@q4a Thank you very much!

Nothing quite like waking up to see problems already solved :+1:

espkk (commented):
It looks good!
I see you removed some utf8-related code from the original `Dialog` though. Did you mean to drop its support or it's still supported somehow?

PS
@q4a, iirc from our discussion clang is not more strict, it just doesn't support this C++20 feature yet 😅

Hammie:
> I see you removed some utf8-related code from the original `Dialog` though. Did you mean to drop its support or it's still supported somehow?

@espkk I assume you are talking about the modified `AddToStringArrayLimitedByWidth`? I'm confident the new code still works with utf-8, but I will try and find a more conclusive way to make sure.

Hammie:
@espkk I've tried creating a unit test to make sure it still works with non-ascii characters, but the presence of the renderer makes that almost impossible. 

The theory as to why it should still work is that neither ' ' or '\n' could ever by confused for other characters in utf-8 - as is the case for all ascii characters <= code point 127 - because multi-byte code points always have the highest bit active.

Also did a quick test with some Cyrillic text, no idea what it says but it looks like it is rendering correctly:
Sea Dogs 28_08_2022 14_28_55

espkk:
> @espkk I've tried creating a unit test to make sure it still works with non-ascii characters, but the presence of the renderer makes that almost impossible.
> 
> The theory as to why it should still work is that neither ' ' or '\n' could ever by confused for other characters in utf-8 - as is the case for all ascii characters <= code point 127 - because multi-byte code points always have the highest bit active.
> 
> Also did a quick test with some Cyrillic text, no idea what it says but it looks like it is rendering correctly

Sounds good! Thanks

But this needs a rebase now 😢

espkk (dismissed):
only 1 minor note - please move new header ito `src` if it's not public
`include` is for public headers which are getting exported

espkk (approved):

PR #386, opened 2022-07-22
Update ship texture folder for PotC ships

Hammie:
Makes the game look for ship textures in the `Ships/` folder, for compatibility with PotC & New Horizons.


espkk:
Is there any specific reason why we don't want doing that from scripts? 
`SetTexturePath` script function supports up to 4 levels for lookup directories. The implementation is really ugly but it works and used in different versions of the game...

Hammie:
> Is there any specific reason why we don't want doing that from scripts? `SetTexturePath` script function supports up to 4 levels for lookup directories. The implementation is really ugly but it works and used in different versions of the game...

I didn't know that was exposed in the scripts. I'll see if I can get it to work that way.

Hammie:
Script changes did manage to fix it. So closing this.

PR #388, opened 2022-07-22
add clang-12 build to Linux CI

q4a:
It's usefull to have 2 build versions: clang-11 - for prebuild conan packages and clang-12 for building packages from source code.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -34,7 +34,8 @@ jobs:
       fail-fast: false
       matrix:
         configuration: [Debug]
-    name: 'build-linux [${{ matrix.configuration}}]'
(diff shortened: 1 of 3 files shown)


Hammie (approved):
Looks good to me

Hammie:
@q4a Is the CI stuck?

q4a:
Yes, but I don't know why. It's Windows only, but changes related to Linux only. That's why I think: it's not problem.

q4a:
That's happens when I remove wayland dependency: https://github.com/q4a/storm-engine/runs/7481875101?check_suite_focus=true
Same error if I just remove wayland recipe hash: https://github.com/q4a/storm-engine/runs/7481972287?check_suite_focus=true#step:5:10447

PR #389, opened 2022-07-22
Make flag textures dynamically configurable

Hammie:
Makes flag textures configurable at runtime by setting the `texture`, `textureColumns` and `textureRows` attributes.
Adds a MSG_SHIP_SET_CUSTOM_FLAG message that enables you to use a specific flag entity per ship.

Combined these enable the use of a large amount of different flags spread across different textures.

### DIFF
--- a/src/libs/rigging/src/flag.cpp
+++ b/src/libs/rigging/src/flag.cpp
@@ -13,7 +13,6 @@ FLAG::FLAG()
 {
     bUse = false;
     RenderService = nullptr;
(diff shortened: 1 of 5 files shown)


Hammie:
Also added the ability to disable individual flags by returning `-1` from `GetRiggingData` (row or column).

espkk (approved):
Looks good! Is this ready for merge?

Hammie:
Yes, it can be merged. Did some more testing and everything is working great now.

PR #391, opened 2022-07-23
Fix SegmentIsLoaded always returning false

Hammie:
`SegmentIsLoaded` was always returning false because the file name string was overridden (empty string) before being used


espkk (approved):
Thanks!
I'm wondering if this is a regression from the stock version...

Hammie:
The code itself did not change compared to Storm 2.0. But I think there are some small changes to the way the scripting stack behaves.

PR #395, opened 2022-08-02
Transferable ships

ann0yer:


Hammie:
Code looks good.

I am wondering if it might be better to do this through a `"transferable"` attribute on `pShipAttr` instead?

ann0yer:
> Code looks good.
> I am wondering if it might be better to do this through a "transferable" attribute on pShipAttr instead?

Thank you for review.
I'm not sure; in fact I just separted "transferable" responsibility from the `isMyShip` flag, so implementing it in the same way looked logical to me. Are you worried about `BI_IN_CREATE_SHIP` message signature becoming too various?

Hammie:
> > Code looks good.
> > I am wondering if it might be better to do this through a "transferable" attribute on pShipAttr instead?
> 
> Thank you for review. I'm not sure; in fact I just separted "transferable" responsibility from the `isMyShip` flag, so implementing it in the same way looked logical to me. Are you worried about `BI_IN_CREATE_SHIP` message signature becoming too various?

I prefer doing things through attributes where possible, but I've also had another look at the code and adding this to ship-stats would also not make much sense.

So I agree this is the best approach right now.

Hammie (approved):

PR #396, opened 2022-08-02
Fix crash when creating AIGroup during SituationCheck

Hammie:
New Horizons has a feature where ships can surrender during battle. This check happens during the `SHIP_CHECK_SITUATION` event. This creates a new "surrendered" group, potentially resizing the static `AIGroup::AIGroups` vector, which then invalidates the iterators in `SEA_AI::Execute` it is called from, crashing the game.

This is a workaround by simply using the index instead of an iterator in the loop.


espkk (approved):
Interesting...I believe this was clang-tidy auto conversion
Ideally, we need to try to get rid of such
But as a fix it looks good. Thank you!

PR #397, opened 2022-08-03
Add options for display selection and window borders to engine.ini

Hammie:
Exposes two new options to the engine.ini file:

`display`: the index of the display/monitor the game should be shown on, default = `0`
`window_borders`: show window borders when running in windowed mode, allowing you to move the window, default = `0` / disabled


ugeen4 (approved):
Looks good !

PR #399, opened 2022-08-06
Fix rendering glitch caused by XI_Glower

Hammie:
`DX9RENDER::DrawRects` resets the world matrix without restoring it afterwards.

This caused an issue in combination with the `CXI_GLOWER` interface element, causing the world matrix to be reset during mouse cursor drawing.

I'm pretty sure none of the newer games used the _Glows_ element, so that is probably why this went undetected.

### DIFF
--- a/src/libs/renderer/src/s_device.cpp
+++ b/src/libs/renderer/src/s_device.cpp
@@ -3394,8 +3394,9 @@ void DX9RENDER::DrawRects(RS_RECT *pRSR, uint32_t dwRectsNum, const char *cBlock
 
     bool bDraw = true;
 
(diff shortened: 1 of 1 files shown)


espkk (approved):
I'm wondering why this affects only `XI_Glower`, since DrawRects is used in many places without restoring the world matrix
Anyway, it's good if this fixes the issue

Hammie:
I think its the only xinterface element to use `DrawRects`, as it is a sort of UI-layer particle effect.

On top of that, maybe the rest of the non-xinterface code also uses an identity world matrix, or maybe sets their own every time.

PR #403, opened 2022-08-16
Gamma/brightness/contrast in windowed mode

espkk:
mimalloc-redirect is disabled in this PR because of the windows bug: it uses RtlFreeHeap after allocating some resource with malloc which is detoured by mimalloc-redirect. This results in a crash e.g. if  `SetGammaRamp ` or `SetDeviceGammaRamp` is used.
I hope this will be workarounded by mimalloc or fixed by microsoft

### DIFF
--- a/src/apps/engine/CMakeLists.txt
+++ b/src/apps/engine/CMakeLists.txt
@@ -1,10 +1,10 @@
 # this is something I really don't like but it seems there's not better way to ensure mimalloc is linked before c-runtime
-include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)
-if (WIN32)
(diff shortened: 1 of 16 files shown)


Hammie:
What does `mimalloc-redirect` actually do? And what does it mean if we disable it?

espkk:
> What does `mimalloc-redirect` actually do? 

it hooks CRT memory management redirecting them to mimalloc. It looks like a function in GDI allocates memory with CRT and then deallocates using some different API. so there is an allocator mismatch that causes crashes

> And what does it mean if we disable it?

external DLLs will use UCRT allocator -> performance may degrade. but I don't have any real-life benchmarks

ugeen4 (approved):
tested on latest TEHO version, seems to be ok!

PR #404, opened 2022-08-19
Remove common library

espkk:
- move headers into corresponding modules for the sake of maintainability and compile times
- remove some unused stuff
- modernize some rudiments

(no functional changes should've been made)

PS the project structure still sorely lacks of private/public deps. but this is not the scope of this change

PPS I would expect Linux build to be broken


q4a:
Works fine on Linux. Thanks!

Hammie (approved):
This is a large chunk of changes, but a very welcome one!

Also quickly tested it with TEHO, and everything seems to work perfectly.

PR #406, opened 2022-08-24
bring back sabergun

Artess999:


src/libs/location/src/character.cpp
```diff
@@ -3224,6 +3228,13 @@ bool Character::zSetBlade(MESSAGE &message)
         isBladeSet = false;
         SetFightMode(false);
     }
+
+    // sabergun
+    if (nBladeIdx == 1 && !name.empty())
+    {
+        isGunSet = true;
```
espkk: otherwise set to `false` (in case regular blade will be loaded afterwards)

espkk (commented):
Looks good, ty!
please take look at mentioned point

src/libs/location/src/character.cpp
Artess999: Thanks! Fixed

espkk (approved):

PR #407, opened 2022-08-27
[xinterface] fix editing text button pos

espkk:


espkk:
it turns out this basically reverts #236
no idea what was its purpose. @mitrokosta?

espkk:
oh well, is it affected for setting from scripts? then it should be solved in some other way because #236 breaks the interface editor

PR #410, opened 2022-08-28
[script_defines]: uncomment RELATION_ALLIANCE

Artess999:
Is used in AoP: Caribbean Tales scripts

(Changed to 4 because 3 is already used in RELATION_UNKNOWN)

### DIFF
--- a/src/libs/shared_headers/include/shared/sea_ai/script_defines.h
+++ b/src/libs/shared_headers/include/shared/sea_ai/script_defines.h
@@ -89,11 +89,11 @@
 // AI Section
 
 // Nation section
-#define RELATION_UNKNOWN 3
-#define RELATION_FRIEND 0
-#define RELATION_NEUTRAL 1
-#define RELATION_ENEMY 2
-//#define RELATION_ALLIANCE        3
+#define RELATION_UNKNOWN 0
+#define RELATION_FRIEND 1
+#define RELATION_NEUTRAL 2
+#define RELATION_ENEMY 3
+#define RELATION_ALLIANCE 4
 
 #define AI_ALL 10000
 #define AI_SELECT 10010


Hammie (dismissed):

espkk:
What if we put unknown to 0 and shift everything? Otherwise it looks ugly..

Artess999:
> What if we put unknown to 0 and shift everything? Otherwise it looks ugly..

Makes sense... Fixed, thanks!

espkk:
it will also probably crash here:
https://github.com/storm-devs/storm-engine/blob/1bed196dea514c9173400e097357f4f2f55fd177/src/libs/sea_ai/src/ai_ship_camera_controller.cpp#L182
maybe we need to extend this array. If there's no specific color attribute for `alliance` we can set it to `friend`...

espkk:
and also it will probably break existing saves...

Artess999:
> it will also probably crash here:
> 
> https://github.com/storm-devs/storm-engine/blob/1bed196dea514c9173400e097357f4f2f55fd177/src/libs/sea_ai/src/ai_ship_camera_controller.cpp#L182
> 
> 
> maybe we need to extend this array. If there's no specific color attribute for `alliance` we can set it to `friend`...



Guess it's not needed :)
Alliance is about countries, not characters

Снимок экрана 2022-08-31 в 21 41 25

Also there's only 4 colors in AoP exactly because of this
Снимок экрана 2022-08-31 в 21 43 04

espkk:
@Artess999, well, -1 should've also been working at that point... But we had crashes :)
Anyway it only concerns addons that use `RELATION_ALLIANCE`, so this is definitely better than nothing

espkk (approved):

PR #411, opened 2022-09-17
Simplify and cleanup FONT code

Hammie:
Removed some unused code, added some extra parameters to `Print` and `GetStringWidth` so we can get rid of `(Re)storeFontParameters`, and did some general modernization and renaming to make it more consistent with our coding guidelines.

Functionally, nothing has been changed.

### DIFF
--- a/src/libs/renderer/src/font.cpp
+++ b/src/libs/renderer/src/font.cpp
@@ -2,48 +2,24 @@
 #include "core.h"
 #include "utf8.h"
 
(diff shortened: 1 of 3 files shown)


espkk (approved):
Looks cool! Have you tested all this for regressions?

Hammie:
@espkk Tested it with TEHO and New Horizons, did not notice any differences or issues.

PR #412, opened 2022-09-17
[legacy-dialog]: add sound

Artess999:


src/libs/dialog/src/legacy_dialog.cpp
```diff
@@ -409,6 +431,26 @@ void LegacyDialog::DrawBackground(size_t start, size_t count)
...
+
+    const auto model = dynamic_cast<MODEL *>(core.GetEntityPointer(headModel_));
+
+    model->GetAnimation()->CopyPlayerState(0, 1);
+
+    model->GetAnimation()->Player(0).SetAction(action.c_str());
+    model->GetAnimation()->Player(0).Play();
+
+    model->GetAnimation()->Timer(0).ResetTimer();
+    model->GetAnimation()->Timer(0).Start(0.2f);
```
Hammie, line 459: Why `0.2f`?

Artess999: This value is taken from original code
Without it animationt don't changing

Hammie (approved):

espkk:
Hold this a bit until I find out how to deal with broken sound slots

espkk:
#413 should fix `SoundIsPlaying` issue

PR #415, opened 2022-09-19
Dodge

EvgAnat:


espkk:
Thanks for the PR!
Can you please do clang-format and remove meaningless comments? We use git history for that

EvgAnat:
Thanks for comment. I tried to fix these issues. Can you check the code once again?

src/libs/location/src/character.cpp
```diff
@@ -4938,8 +5081,8 @@ inline void Character::CheckAttackHit()
             fc.c->Attack(this, fgt_attack_feintc);
             fc.c->UpdateAnimation(); // boal
             // fc.c->PlaySound("fgt_feint");
-        }
-        else
+        }
```
espkk: space here

src/libs/location/src/character.cpp
```diff
@@ -2828,21 +2931,48 @@ if(storm::iEquals(eventName, "Blade to belt"))
...
                 else if (storm::iEquals(eventName, CHARACTER_FIGHT_GUNFIRE))
                 {
+                    isFireState = false;
                     core.Send_Message(blade, "l", MSG_BLADE_GUNFIRE);
                     // PlaySound("pistol_shot");
                     isFired = true;
                     float kDist;
                     Character *chr = FindGunTarget(kDist, CheckShotOnlyEnemyTest());
                     entid_t enemy{};
+                    int32_t resHit = 0; // EvgAnat: 0 - null target, 1 - miss, 2 - hit
```
espkk: comment here

src/libs/location/src/character.cpp
```diff
@@ -4482,7 +4613,7 @@ void Character::UpdateAnimation()
                 case fgt_strafe_l: // Bounce to the left
                     core.Send_Message(blade, "ll", MSG_BLADE_TRACE_OFF, 0);
                     recoilSound = SOUND_INVALID_ID; // PlaySound("recoil", true);
-                    impulse += 15.0f * CVECTOR(-cosf(ay), 0.0f, sinf(ay));
+                    impulse += 10.0f * CVECTOR(-cosf(ay), 0.0f, sinf(ay));
```
espkk: what's the reason for changing the original value? if it's required, this should be configured from scripts

EvgAnat: 15.0 in game looks grotesque, especially when NPCs often choose dodging by strafe. Empirically 10.0 is fine.

src/libs/location/src/character.cpp
```diff
@@ -2828,21 +2931,48 @@ if(storm::iEquals(eventName, "Blade to belt"))
...
                 else if (storm::iEquals(eventName, CHARACTER_FIGHT_GUNFIRE))
                 {
+                    isFireState = false;
                     core.Send_Message(blade, "l", MSG_BLADE_GUNFIRE);
                     // PlaySound("pistol_shot");
                     isFired = true;
                     float kDist;
                     Character *chr = FindGunTarget(kDist, CheckShotOnlyEnemyTest());
                     entid_t enemy{};
+                    int32_t resHit = 0;
```
Artess999: I guess it might be a good idea to use enum instead of simple values
@espkk what do you think about it?

src/libs/location/src/character.cpp
Artess999: Agreed with @espkk, this values should be configured from scripts
If nothing sent from scripts, than vanila values should be used

src/libs/location/src/character.cpp
```diff
@@ -2828,21 +2931,48 @@ if(storm::iEquals(eventName, "Blade to belt"))
...
-                        chr->Hit(fgt_hit_fire);
+			resHit = 1;
+			VDATA *vd = core.Event("Check_ChrHitFire", "iilf", GetId(), enemy, static_cast<int32_t>(chr->isRecoilState), kDist);
+			if(vd)
+			{
+                            if(vd->Get(resHit))
+                            {
+                                if(resHit == 2)
+                                {
+                                    core.Event("Location_CharacterFire", "iifl", GetId(), enemy, kDist, 1);
```
Artess999: smells bad
`core.Event("Location_CharacterFire", "iifl", GetId(), enemy, kDist, ...);` is duplcated four times
Maybe add variable with last value and only change it?

therefore you could bring back event you've deleted, but it'll look like
`core.Event("Location_CharacterFire", "iifl", GetId(), enemy, kDist, myNewAwesomeVariable);`
Also in this case last `else` on 2974 line could be removed

EvgAnat: Sure. That was a little spaghetti because I didn't adapt my old dodge well for compatibility. Also "jedi bullet" was from here.

src/libs/location/src/character.cpp
EvgAnat: Fixed. Now resHit is a standard boolean with {0; 1} value.

src/libs/location/src/character.cpp
EvgAnat: Fixed.

src/libs/location/src/character.cpp
EvgAnat: Fixed.

src/libs/location/src/character.cpp
Artess999: Maybe then it should be boolean, not int?

Artess999: also I suppose that `isEnemyFound` sounds better than `resHit` and easier to understand what is happening here

Artess999: https://github.com/storm-devs/storm-engine/blob/f6c8caad16b4126257a082d53345d3673b5100f8/src/libs/location/src/character.cpp#L4981
https://github.com/storm-devs/storm-engine/blob/f6c8caad16b4126257a082d53345d3673b5100f8/src/libs/location/src/character.cpp#L4982

Check this out
`isGunBlade` here is bool

Artess999: or even add two bools for better readability
`isEnemyFound` for `Location_CharacterFire` and `isEnemyHitFire` for `Check_ChrHitFire `

EvgAnat: > or even add two bools for better readability
`isEnemyFound` for `Location_CharacterFire` and `isEnemyHitFire` for `Check_ChrHitFire`

What is the first bool?
Btw I use int to get boolean from script according to the general codebase.
Like `tmpBool` in this function, for example:
image

Artess999: Hmmmm, nevermind about bool
I'm suggesting something like this:

```
                    entid_t enemy{};
                    int32_t isEnemyHitByGunfire = 1;

                    if (chr)
                    {
                        enemy = chr->GetId();
                        VDATA *vd = core.Event("Check_ChrHitFire", "iilf", GetId(), enemy,
                                               static_cast<int32_t>(chr->isRecoilState), kDist);
                        if (vd)
                        {
                            vd->Get(isEnemyHitByGunfire);
                        }

                        if (isEnemyHitByGunfire)
                        {
                            chr->Hit(fgt_hit_fire);
                        }
                    }
                    core.Event("Location_CharacterFire", "iifl", GetId(), enemy, kDist,
                               chr != nullptr && isEnemyHitByGunfire);
```

I think it'll be more clear for future reader : )

src/libs/location/src/character.cpp
```diff
@@ -1621,10 +1623,25 @@ void Character::Recoil()
         return;
     if (!GetLocation()->IsSwimming())
         return;
-    fgtSetType = fgt_recoil;
-    fgtSetIndex = 0;
-    recoilWait = 0.8f;
-    isTurnLock = false;
+    VDATA *res = nullptr;
+    res = core.Event("ChrCheckEnergy", "is", GetId(), "recoil");
+    int32_t isEnable = 1;
```
Artess999: `isEnabled` : )
also what is enabled? this is not obvious name

EvgAnat: Corsairs coding makes my English worse :)
`isEnabled` signals that current action is possible. Passed the energy test or is enabled by default. `isEnoughEnergy` isn't actual if event doesn't exist in scripts. Maybe `isActionPossible`?

src/libs/location/src/character.cpp
EvgAnat: I don't like operand `chr != nullptr`. You guess it's necessary?

src/libs/location/src/character.cpp
EvgAnat: Fixed.

src/libs/location/src/character.cpp
Artess999: Yep, I think so

src/libs/location/src/character.cpp
Artess999: I like `isActionPossible`

EvgAnat: Done.

src/libs/location/src/character.cpp
EvgAnat: Done.

src/libs/location/src/np_character.cpp
```diff
@@ -858,11 +858,62 @@ void NPCharacter::DoFightActionAnalysisNone(float dltTime, NPCharacter *enemy)
...
+                    auto chr2 = static_cast<NPCharacter *>(fch.c);
+                    if (chr2 == this)
+                    {
+                        isEnemyFire = true;
+                        break;
+                    }
+                }
+            }
+        }
+    }
```
espkk: what is the meaning of this check?

EvgAnat: Npc checks other characters in sector: is there at least one enemy who aiming a gun at this npc?

espkk: This may do a lot of calculations in some situations. Ideally, this should be done in reverse, marking the aim target.
But since it would involve more complex logic it probably makes sense to at least make this execution happen only if supported by scripts. e.g. by calling dodge probability event before this clause and checking if it was processed
Other than that LGTM

EvgAnat: Agree. I added another one check.

espkk (approved):

PR #417, opened 2022-09-24
[worldmap] fix high fps glitches

espkk:
There was a value used for vertex pos calculation which increased in geometric progression. If fps was high enough this caused glitches similar to this


Hammie (approved):

Hammie:
What game/mod is that screenshot from?

Artess999:
> What game/mod is that screenshot from?

CorsairsShipPack

This bug also appears in POTC fork with `ondebugkeys = 1` and pressed R key (x10 timescale)

espkk:
> What game/mod is that screenshot from?

this is a COAS mod mentioned above. currently 99% of bug reports come from its players

PR #418, opened 2022-09-25
Simplify string_codec

Hammie:


espkk:
I like the idea of simplifying that ancient stuff. I'm also wondering can we just replace that STRING_CODEC with some fast modern hash table with heterogenous lookup support? 

there are some nice fresh benchmarks: https://www.reddit.com/r/cpp/comments/x82kac/comprehensive_c_hashmap_benchmarks_2022/

Hammie:
Yeah, after simplifying it, it becomes quite clear there is not much left. I'd be happy to try out a good hash map instead, only really worked with `std::unordered_map` in the past myself.

I did already do some doodles with `std::unordered_map` before, to mixed success (I think I just had a few mistakes somewhere in my code).

Did learn that the hash codes are also saved with the save games, so we either need to make sure the hashing algorithm stays the same or make some migration code to deal with old save games.

espkk:
> Yeah, after simplifying it, it becomes quite clear there is not much left. I'd be happy to try out a good hash map instead, only really worked with `std::unordered_map` in the past myself.
> 
> I did already do some doodles with `std::unordered_map` before, to mixed success (I think I just had a few mistakes somewhere in my code).
> 
> Did learn that the hash codes are also saved with the save games, so we either need to make sure the hashing algorithm stays the same or make some migration code to deal with old save games.

We just couldn't do this properly with std::unordered_map because heterogeneous lookup (`is_transparent` stuff) was not supported prior to C++20. Now we can have something like this: 
`std::unordered_map<Hash, ATTRIBUTE> map;`

where `Hash` plays the role of "string codec" and accepts any of std::string or string_view or const char* or already precalculated hash.  While calculating it updates its own std::unordered_map<HASH, std::string> that's being saved/loaded from saved state data.

We only need to make sure that all the hashing stuff that's currently spread across the engine(vma,core_impl,vstring_codec, maybe somewhere else) is in the same place, e.g. in `util/hash.hpp`.e 
We can keep the original hashing algorithm, but since it's not the only recent change that requires a new game it probably makes sense to pick a better algorithm... we only need to wrap it into case-insensitive loop.

This would also allow us to turn some `uint32_t hash` hash into `Hash hash` in order to precalculate some hashes at the compile time.

UPD: hmm I think we don't even need heterogenous lookup since we will use our own hash:string storage...

espkk:
I don't mind this change anyway, but have to review it more thoroughly since we've already stepped on this rake with `string_view`...

src/libs/core/src/string_codec.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+class STRING_CODEC final : public VSTRING_CODEC
+{
+  public:
+    STRING_CODEC();
+    ~STRING_CODEC() noexcept override;
+
+    [[nodiscard]] uint32_t GetNum() const override;
+    uint32_t Convert(const std::string_view &pString) override;
+    [[nodiscard]] std::string_view Convert(uint32_t code) const override;
```
espkk, line 26: Isn't this UB? I think we are relying on the behavior when all strings are null-terminated. While if we store std::string and return std::string_view to it there is no such guarantee

espkk: It probably can be hacked by constructing a `string_view` from `{ string.c_str(), string.c_str()+string.length() }`
In the real world, I'd expect both to produce the same result always, but formally only call to c_str() guarantees to return a null-terminated string

Hammie: Good point, didn't really think of that

Hammie: I've gone through and updated all the affected code to handle string_views correctly, either using the string_view directly or converting to a normal string as needed.

I now think `string_view.data()` should probably be considered a code smell.

espkk (approved):

PR #419, opened 2022-09-27
sea_ai: support HeightMultiply for balls

espkk:
logical continuation of #321 

- originally height was calculated incorrectly, making HeightMultiply unusable
- in stock COAS there is a cannon with a custom multiplier=0.6; this also should fix trajectories for it

### DIFF
--- a/src/libs/sea_ai/src/ai_balls.cpp
+++ b/src/libs/sea_ai/src/ai_balls.cpp
@@ -3,12 +3,11 @@
 #include "math_inlines.h"
 #include "shared/messages.h"
 
-
 AIBalls *AIBalls::pAIBalls = nullptr;
 
 AIBalls::AIBalls()
-    : pSail(nullptr), pSea(nullptr), pFort(nullptr), pIsland(nullptr),
-      fDeltaTimeMultiplier(1.0f), fBallFlySoundDistance(1.0f), dwFireBallFromCameraTime(0)
+    : pSail(nullptr), pSea(nullptr), pFort(nullptr), pIsland(nullptr), fDeltaTimeMultiplier(1.0f),
+      fBallFlySoundDistance(1.0f), dwFireBallFromCameraTime(0)
 {
     pAIBalls = this;
 }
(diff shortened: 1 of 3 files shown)


espkk:
This modiifes `CannFire` event arg
`fDirAng = GetEventData();`
that was not used in any game/mod

Now it is `fRawAng` which should be forwarded to `AIBalls` like this (e.g. in `Ball_AddBall`):
`AIBalls.RawAng = fRawAng;`

Hammie:
New Horizons seems to be using the `fDirAng`, so I'm not sure about removing that completely.

Will try and figure out why its being used, but the script looks quite confusing.

espkk:
I can introduce additional argument, but I was sure that angle is not used since it seems to be completely pointless...
Could you share that code?

Hammie:
> I can introduce additional argument, but I was sure that angle is not used since it seems to be completely pointless... Could you share that code?

It seems to be used for determining what from which quadrant it is being fired, so they can look up additional properties of the cannons/ships.

```
if(USE_REAL_CANNONS) {
	float ourdir = clampangle(fDirAng);
	...
	if(aCharacter.ship.type != SHIP_FORT_NAME) {
		int quad = quadrantG(ourdir, clampangle(fCannonDirAng));
		...
		string quadstr = GetQuadName(quad);
		...
	}
	...
}
```

with 

```
int quadrantG(float ship, float ball)
{
	ship = clampangle(ship); ball = clampangle(ball);
	// rotate angles so ship is 0
	ball = ball - ship; // + PI/4.0; // so that it's not +-45degrees, it's 0-90deg
	//ship = 0;
	//do it here.
	//trace("Doing check. Ball = " + Radian2Degree(ball) + ", Ship = " + Radian2Degree(ship));
	int ang = round(Radian2Degree(ball));
	while (ang >= 360) ang -= 360;
	while(ang < 0) ang += 360;
	if(ang < 180+30)
	{
		if(ang < 30) return 0;
		if(ang < 180-30) return 1;
		return 2;
	}
	else
	{
		if(ang < 360-30) return 3;
		return 0;
	}
	//return quadrant(ball, 0.075);
	/*if(ship > PI)
	{
		tmp = PIm2 - ship;
		ball += tmp;
	}
	return quadrant2(ang1+PI/2.0, ang2+PI/2.0);*/
}
```

espkk:
@Hammie , so it basically detects shipboard by calculating the angle for each shot?? I think it can just be taken from `SHIP_BORT_FIRE` event...
anyway, I'll add additional param

espkk:
> @Hammie , so it basically detects shipboard by calculating the angle for each shot?? I think it can just be taken from `SHIP_BORT_FIRE` event... anyway, I'll add additional param

nvm, I forgot about randomized fire delay.

PR #421, opened 2022-09-27
Fix edit mode not working in dialogs

Hammie:
This fixes a regression caused by my dialog refactoring.

Edit mode starts of with an empty dialog entry, which caused it to skip that entry, making the edit field disappear.


espkk:
Nice, thanks! 
I'll ask bug reporter to confirm it works

espkk (approved):

PR #425, opened 2022-10-03
[core] rework script cache

espkk:
- invalidation depends on folder content timestamp instead of corresponding file hash
- generic refactoring and bug fixing
- `use_cache{true|false}` changed to `cache_mode{0|1|2}` 0 - disabled
1 - enabled, integrity calculated on each script load 2 - enabled, integrity calculated once on start

'1' recommended for development, e.g. interface script tuning 
'2' recommended for release

### DIFF
--- a/src/libs/core/src/compiler.cpp
+++ b/src/libs/core/src/compiler.cpp
@@ -1,7 +1,7 @@
 #include "compiler.h"
 
-#include <cstdio>
 #include <chrono>
+#include <cstdio>
 
 #include <zlib.h>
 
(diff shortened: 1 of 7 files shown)


src/libs/core/src/compiler.cpp
```diff
@@ -405,7 +459,11 @@ void COMPILER::LoadPreprocess()
         else
             bRuntimeLog = true;
 
-        use_script_cache_ = engine_ini->GetInt("script", "use_cache", false);
```
Hammie, line 408: We could still use the `use_cache` value as a fallback for when `cache_mode` is not defined, but since scripting caching is a fairly new feature (I think?) I don't think backwards-compatibility is important.

Hammie (approved):

PR #427, opened 2022-10-06
Legendary Edition changes. Part 1

q4a:
This changes was made by @ugeen4 and he allowed me to create PR for current develop branch.
This is 1st part of changes, which will not change default TEHO behavior.


Hammie (commented):
If I understand correctly this:

- Makes the max/min resolution of the Sea mesh configurable
- Adds an optional line spacing multiplier for table UI
- Makes font colors configurable for selected/pressed states
- Makes the line height configurable for text in image scrolls
- Add support for moving the cursor right in text inputs (a bugfix?)

If you can confirm I understand it correctly I think it can be merged

q4a:
@Hammie Yes, you understand is correct.

Hammie (approved):

PR #433, opened 2022-10-23
Improve slide_line behavior

q4a:
Current slide_line behavior is strange: if you click 1 time on arrow on slide_line - it will chnange position to min (0) or max (m_nGrateQuantity):
old

@ugeen4 improved slide_line behavior in Legendary Edition. I think it works better now and I would like to use it in TEHO:
new


espkk:
LGTM, thanks

espkk (approved):

PR #435, opened 2022-11-04
worldmap: simplify calculation of food and rum position in wdm_wind_ui

q4a:
`rum` is `std::optional<int32_t>` and when `rum == 0` we still has `if (rum)` - true. So there was 2 options: remove this check or replace it with `if (rum != 0)`, but if i'll use `if (rum != 0)` - i get this behavior:
rum-0-wrong
Which is wrong and looks bad.
I checked if it's possible to hide rum icon, but it's grouped with few other
(check `resource\Textures\WorldMap\Interfaces\MORALE.tga.tx`):
ConvTX

So, I suggest to remove 2 useless checks.

### DIFF
--- a/src/libs/worldmap/src/wdm_wind_ui.cpp
+++ b/src/libs/worldmap/src/wdm_wind_ui.cpp
@@ -277,7 +277,7 @@ void WdmWindUI::LRender(VDX9RENDER *rs)
     DrawRects(buf, 1, "WdmDrawMapBlend");
 
     // write the number of supplies
-    float foodRumSpacing = rum ? 24.0f : 0.0f;
+    float foodRumSpacing = 24.0f;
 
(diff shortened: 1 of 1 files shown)


ugeen4 (approved):

Hammie (changes requested):
I think I added that check because not every game has rum.

So they are not useless, and a better "fix" would be to make those icons more dynamic as well.

espkk:
I think it's intended behavior
What's the issue btw, I didn't get it...

Artess999:
There's no rum on worldmap in AoP and CoAS : )
In PotC too...

image
image

q4a:
Sorry, I didn't check file history - when this checks appeared. There is no error or issue. I just want to simplify checks.

PR #436, opened 2022-11-05
Sea foam optimization

ann0yer:


espkk:
Thanks for that finding! Fixed in #456, but probably this can be adjusted to provide further optimizations/refactoring

PR #438, opened 2022-11-06
Fix worldmap setting an invalid encounter type

Hammie:
For the `WM_UpdateCurrentAction` event, the worldmap was setting an incorrect value for "no encounter".

It was trying to set it to `-1`, which was then interpreted as an unsigned integer and converted to `4294967295`. In the scripts (`WM_UpdateCurrentAction`), this value is then converted back to an integer using `sti`. But because `4294967295` is outside the range of a signed integer it instead returned a default value of zero `0`, causing the script to think there is always an encounter even when there was none.

I've replaced the `-1` with another magic value that does work (int32-max). And also added all encounter magic values to the shared events header (is there a more appropriate header?) for ease of use.

@ugeen4 @q4a I've noticed that in TEHO, you are never shown "near land". Not sure if this was an intentional decision or the result of this bug, but if it is the latter this should allow you to fix that.


src/libs/shared_headers/include/shared/events.h
```diff
@@ -19,4 +19,12 @@
 // crosshair
 #define TELESCOPE_ACTIVE "TelescopeActive"
 
+// Worldmap
+#define WORLDMAP_ENCOUNTER_TYPE_UNKNOWN 0
+#define WORLDMAP_ENCOUNTER_TYPE_MERCHANT 1
+#define WORLDMAP_ENCOUNTER_TYPE_WARRING 2
+#define WORLDMAP_ENCOUNTER_TYPE_FOLLOW 3
+#define WORLDMAP_ENCOUNTER_TYPE_STORM 4
+#define WORLDMAP_ENCOUNTER_TYPE_INVALID 2147483647
```
espkk: do the scripts support hex literals? if so, `0x7ffffff` would be much more self-explanatory...

Hammie: I've also added a small comment for anyone who might be print-debugging this

ugeen4 (dismissed):

src/libs/shared_headers/include/shared/events.h
Hammie: Turns out scripts don't support hex numbers, and it's very important to re-copy these headers when testing this.

espkk (approved):

PR #443, opened 2022-11-27
Use SDL_GetWindowSize and SDL_WarpMouseInWindow in Windows

q4a:
This is mostly about using SDL_GetWindowSize and SDL_WarpMouseInWindow in Windows
and few include changes.

### DIFF
--- a/src/libs/window/include/os_window.hpp
+++ b/src/libs/window/include/os_window.hpp
@@ -6,6 +6,12 @@
 
 namespace storm
 {
+struct WindowSize
+{
+    int width{};
+    int height{};
+};
+
 //! Abstract window
 class OSWindow
 {
@@ -38,6 +44,8 @@ class OSWindow
     virtual int Width() const = 0;
     //! Current height of window
     virtual int Height() const = 0;
+    //! Current window size
+    virtual WindowSize GetWindowSize() const = 0;
     //! Is window fullscreen
     virtual bool Fullscreen() const = 0;
     //! Window title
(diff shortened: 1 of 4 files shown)


espkk:
IIRC this has been already implemented, but then I reverted it back because it worked just awful on windows due to poorly designed sdl implementation...
Was there any changes since then?

q4a:
> IIRC this has been already implemented, but then I reverted it back because it worked just awful on windows due to poorly designed sdl implementation...

Can you find your revert? What was worked awful? `SDL_GetWindowSize` - already used in `SDLWindow::Width()` and `SDLWindow::Height()` - I can use it.
https://github.com/storm-devs/storm-engine/blob/develop/src/libs/window/src/sdl_window.cpp#L44
`SDL_WarpMouseInWindow` - IDK, here is source code:
https://github.com/libsdl-org/SDL/blob/0a48abc860b436387c94e90c13fd0db25772e356/src/events/SDL_mouse.c#L1015

q4a:
I splited PR - this will have only SDL changes.
And this one will have all other - https://github.com/storm-devs/storm-engine/pull/448

espkk:
I'm afraid that exposing `SDL_Window` is not an option to go. It ruins the whole point of `OSWindow` abstraction
Instead, you can add window size query to the `OSWindow` interface which can be acquired via core API

q4a:
@espkk sorry for delay. I rebased PR on top of develop and removed exposing SDL_Window from OSWindow class.

src/libs/window/include/os_window.hpp
```diff
@@ -38,6 +38,8 @@ class OSWindow
     virtual int Width() const = 0;
     //! Current height of window
     virtual int Height() const = 0;
+    //! Current window size
+    virtual void WindowSize(int *width, int *height) const = 0;
```
espkk: Could you return a tuple or pair or struct instead of args?

q4a: Changed to struct.

Hammie (approved):
I also did a local merge and it seems to work well on Windows.

PR #447, opened 2022-11-30
Revert "Simplify string_codec"

espkk:
Reverts storm-devs/storm-engine#418

For some reason it causes many issues of different kinds, corrupting engine state

@Hammie, DM me if you need more info


Hammie (approved):

Hammie:
Will investigate this later, but since it is a non-functional change it can be reverted for now.

PR #450, opened 2022-12-09
core: add search for a suitable aspect ratio for PC_SCREEN interface

q4a:
The search is performed on sections named PC_SCREEN_$W:$H, where $W:$H can be 5:4, 4:3, 16:10, 16:9, 21:9 or other. The aspect ratio of game window may not match exactly, but it should fit in +3%/-3% of the aspect ratio = $W:$H.

If suitable PC_SCREEN_$W:$H was not found, then fallback to default PC_SCREEN section.


ugeen4 (approved):
Tested well on dev-TEHO (LE)

espkk (approved):

Issue #451, opened 2022-12-09
[BUG] Linux build: _DEBUG macro is not set


psmitsu:
**Describe the bug**
When building the project on Linux without any IDE, the `_DEBUG` macro is not set. This prevents some features, in particular, `Breakpoint()` function in game scripts doesn't work. 

**To Reproduce**
1. Build the project as described in [Linux workflow](#https://github.com/storm-devs/storm-engine/blob/develop/.github/workflows/ci_linux.yml)
2. Verify that _DEBUG code has not been compiled
    - In the game's scripts, put `Breakpoint()` at the start of `./PROGRAM/seadogs.c` and run the engine
    - Also can directly check that by looking on the code through a debugger
        - E.g. `disas/m` on `apps/engine/src/main.cpp:main` in `gdb`

**Expected behavior**
As I understand it, debug build should allow for triggering breakpoints through scripts and other features (I think special controls can be set?) that are enabled when `_DEBUG` macro is active.

**Additional context**
AFAIK `_DEBUG` macro is specific to Microsoft Visual Studio.

PR #452, opened 2022-12-09
Enable setting _DEBUG flag on Linux build

psmitsu:
This PR enables setting the _DEBUG macro in CMake when building on Linux. The problem it solves is described in #451

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,9 @@ else()
     # Always generate debug information
     add_compile_options(-g)
     add_link_options(-g)
+
+    # Add _DEBUG flag
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
 endif()
 
 # this is crucial to include libs first


espkk:
@q4a, could you take a look?

q4a:
I tested on Windows 3 `CMAKE_BUILD_TYPE`: `Release`, `Debug` and `RelWithDebInfo`. Visual Studio added `-D_DEBUG` only in `Debug` type. So, it's better to add this lines:
```
    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
        add_definitions(-D_DEBUG)
    endif()
```

psmitsu:
@q4a shouldn't `CMAKE_CXX_FLAGS_DEBUG` only set flags for `CMAKE_BUILD_TYPE=Debug`? I've checked (by `grep`-ing `build.ninja` file) each build, and indeed `-D_DEBUG` flag is present only for `-DCMAKE_BUILD_TYPE=Debug`.

q4a:
@psmitsu Sorry, my mistake. I know that there is `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`, but I newer saw `CMAKE_CXX_FLAGS_DEBUG`. Now I tested and it works fine.
@espkk we can merge it as is.

q4a (approved):

espkk (approved):

PR #454, opened 2022-12-30
Entity manager rework

espkk:
should solve a few issues related to entity lifetime


src/libs/core/src/internal_functions.cpp
```diff
@@ -709,7 +709,7 @@ DATA *COMPILER::BC_CallIntFunction(uint32_t func_code, DATA *&pVResult, uint32_t
         pV2->SetType(VAR_AREFERENCE);
         pV2->SetAReference(core_internal.Entity_GetAttributePointer(ent));
 
-        if (core.GetEntityPointer(ent))
+        if (core.GetEntityPointerSafe(ent))
```
Hammie: Maybe replace these checks that don't do anything with the return value with an `IsEntityValid` check instead?

espkk: It's not exposed but it makes sense...

Hammie (approved):

PR #455, opened 2023-01-01
Expose worldmap label position back as attributes

Hammie:
Makes the worldmap write all label positions it reads from the model file as attributes on the entity.

Enables me to store the locations of islands an towns in the worldmap model without having to duplicate all the data in the code.


espkk:
Can this affect any of existing games?

Hammie:
@espkk as I could tell none of the existing games use the `worldmap.labels.position` attribute, so it should not affect them in any way.

espkk (approved):
Cool, lgtm then, thanks

PR #462, opened 2023-01-25
Validate collision data during load on debug builds

Hammie:
I was debugging some crashes that were caused by invalid collision data.

I've added some code that will do some validation on the collision data when loading models. This can be toggled using a `VALIDATE_COLLISION_DATA` variable. Currently configured to only run this in debug builds.

Hopefully this will make debugging future collision issues easier.

### DIFF
--- a/src/libs/geometry/src/geom_static.cpp
+++ b/src/libs/geometry/src/geom_static.cpp
@@ -9,6 +9,8 @@ Import library main file
 ******************************************************************************/
 #include "geom.h"
 
+#include <fmt/format.h>
+
 #include <cstdint>
 #include <cstring>
 #include <vector>
@@ -17,6 +19,13 @@ Import library main file
 
 namespace
 {
+
+#ifdef _DEBUG
+constexpr bool VALIDATE_COLLISION_DATA = true;
+#else
+constexpr bool VALIDATE_COLLISION_DATA = false;
+#endif
+
 std::vector<uint32_t> getColData(GEOM_SERVICE &srv, const std::string_view &file_name)
 {
     std::vector<uint32_t> result;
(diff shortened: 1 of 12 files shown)


src/libs/geometry/src/geom_static.cpp
```diff
@@ -234,9 +254,9 @@ GEOM::~GEOM()
     if (rhead.flags & FLAGS_BSP_PRESENT)
     {
         // bsp
-        srv.free(btrg);
-        srv.free(vrt);
-        srv.free(sroot);
+        btrg.clear();
```
espkk: no need to call those in dtor

src/libs/model/src/node.cpp
```diff
@@ -280,11 +276,12 @@ NODER::~NODER()
 #endif
 
     delete geo;
-    for (int32_t l = 0; l < nnext; l++)
-        if (next[l] != nullptr)
-            delete next[l];
-    if (nnext > 0)
-        free(next);
+    for (NODE *node : next) {
+        if (node != nullptr) {
```
espkk: probably remove excessive check for clarity

src/libs/geometry/src/geom_static.cpp
```diff
@@ -17,6 +19,13 @@ Import library main file
 
 namespace
 {
+
+#ifdef _DEBUG
```
espkk: maybe 
`#if defined(_DEBUG) || defined(STORM_VALIDATE_COLLISION_DATA)`?
We could have some kind of checked build built in RELEASE but with additional checks like this one. DEBUG builds are so slow that you typically don't want to run them unless you actually want to attach a debugger

src/libs/geometry/src/geom_static.cpp
```diff
@@ -9,6 +9,8 @@ Import library main file
 ******************************************************************************/
 #include "geom.h"
 
+#include <fmt/format.h>
```
espkk: I think we have `<format>` available unless you want some features from the latest fmt version...

espkk (changes requested):
This might be very useful, thanks
Just a few minor points

src/libs/geometry/src/geom_static.cpp
Hammie: Is it finally here? 🥳

src/libs/geometry/src/geom_static.cpp
Hammie: How about putting it under some kind of `SAFE_MODE` flag? That way we can keep adding additional checks without making the configuration needlessly complex.

src/libs/geometry/src/geom_static.cpp
Hammie: Looks like we can't quite use `std::format` for our clang builds?

espkk: Oh..may be
We should definitely update the pipeline

espkk (dismissed):

src/libs/geometry/src/geom_static.cpp
Hammie: I reverted back to using `fmt::format` for now. I suggest updating the pipeline and replace all usages of `fmt::format` later at the same time.

espkk (approved):

PR #463, opened 2023-02-02
linux: replace libsafec with custom inline functions

q4a:
libsafec dependency was removed
List of replaced functions: strcat_s, strcpy_s, strncpy_s

This fixed problem with calling strcat_s(dest, dst_size, source), where dst_size > 4096. This used to happen in AddAttributesStringsToBuffer(...) in file xinterface.cpp on Linux.


espkk:
> This fixed problem with calling strcat_s(dest, dst_size, source), where dst_size > 4096

Why did it cause an issue?

q4a:
> Why did it cause an issue?

I don't know why it cause an issue. Perhaps this is due to the fact that I need both calling: `strcat_s(dest, dst_size, source)` and `strcat_s(dest, source)`. To make is possible, I had to add template + call the internal function `_strcat_s_chk`:
```
#undef strcat_s
errno_t strcat_s(char *restrict dest, rsize_t size, const char *restrict src)
{
    return _strcat_s_chk(dest, size, src, BOS(dest));
}
template <rsize_t size> errno_t strcat_s(char (&dest)[size], const char *src)
{
    return strcat_s(dest, size, src);
}
```

PS issue happed in this line:
https://github.com/storm-devs/storm-engine/blob/aebad4d5402a58291f09beb5b390dc1712155c0c/src/libs/xinterface/src/xinterface.cpp#L3202

q4a:
I created simple test with `libsafec`: https://github.com/q4a/io-test/blob/main/io-test.cpp
And got same issue:
```
IGNORE CONSTRAINT HANDLER: (403) strcat_s: dmax exceeds max
rc: 403
```

espkk:
It turns out `RSIZE_MAX_STR` is 4K by default and should be configured using conan option `strmax`
I'd suggest setting it to at least 32MB

Fun fact: intel's safestringlib has it **hardcoded** to 4K

espkk:
I don't mind merging this since it's only related to Linux build, but keep in mind your implementation is much less safe

espkk (approved):

PR #465, opened 2023-02-20
Fix building with GCC on Linux

q4a:
It's always good to be able to test game on another compiler (somtimes bugs happens in compiler)
In addition, I have not very pleasant experience with debugging with Clang on ARM Linux. GCC woks better in my case.
I'm not sure if we need add GCC to CI. It's more of a nice feature and it's okay if someone breaks it for a short time.

### DIFF
--- a/src/libs/math/include/math3d/color.h
+++ b/src/libs/math/include/math3d/color.h
@@ -18,49 +18,14 @@
 class Color
 {
   public:
-    union {
-        struct
-        {
-            union {
-                struct
-                {
-                    // Red
-                    float r;
-                    // Green
-                    float g;
-                    // Blue
-                    float b;
-                };
-
-                union {
-                    struct
-                    {
-                        // rgb in vector
-                        Vector c;
-                    };
-
-                    struct
-                    {
-                        // rgb in vector
-                        Vector color;
-                    };
-                };
-            };
-
-            union {
-                // Transparency
-                float a;
-                // Transparency
-                float alpha;
-            };
-        };
-
-        struct
-        {
-            // Vector4 representation
-            Vector4 v4;
-        };
-    };
+    // Red
+    float r;
+    // Green
+    float g;
+    // Blue
+    float b;
+    // Transparency
+    float a;
 
     // -----------------------------------------------------------
     // Constructors
@@ -80,8 +45,6 @@ class Color
     Color(const Vector4 &v);
     // Unpack
     Color(uint32_t c);
-    // Unpack
-    Color(int32_t c);
     // Copy constructor
     Color(const Color &c);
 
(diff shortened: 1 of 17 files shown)


src/libs/math/include/math3d/color.h
```diff
@@ -18,49 +18,14 @@
 class Color
 {
   public:
-    union {
```
Hammie, line 21: What was the problem here?

q4a: Current version gives error:
```
src/libs/math/include/math3d/color.h:38:32: error: member ‘Vector Color::<unnamed union>::<unnamed struct>::<unnamed union>::<unnamed union>::<unnamed struct>::c’ with constructor not allowed in anonymous aggregate
   38 |                         Vector c;
      |                                ^
```

GCC allows only:
https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html
And GCC's implementation of the feature does not allow the unnamed struct to have members that have non-trivial constructors.

So this works in GCC:
```
union {
    struct
    {
        // Red
        float r;
        // Green
        float g;
        // Blue
        float b;
        // Transparency
        float a;
    };

    // Vector4 representation
    Vector4 v4;
};
```

And this is not:
```
union {
    struct
    {
        // rgb in vector
        Vector color;
        // Transparency
        float a;
    };

    // Vector4 representation
    Vector4 v4;
};
```

espkk: That doesn't sound to be the case. Both classes have trivial(empty) constructors precisely for that reason.
Can you make a godbolt snippet?

q4a: @espkk here you are: https://godbolt.org/z/8468cnoPs

espkk: Move `default`ing into the class, that should work. I remember doing that, probably I missed it for Vector4

espkk: IIRC, it's also possible to make a constructor inside the union to make default initialization explicit. I would avoid something like that, but the approach with `default` constructors should be totally fine and so far I don't see any non-standard or UB there

q4a: @espkk ok. How to fix same error with non default constructor?
https://godbolt.org/z/caMfoEjTT

q4a: @espkk I'm not sure how to do that. I tried `Vector color();` and `Vector color = new Vector();` - but it's not working: https://godbolt.org/z/aP5fqcjKd
Moving constructors into the class is not helping too for other constructors: https://godbolt.org/z/jaE3fz3dv

espkk: Oh, gcc is so "special"
That one should work, but I'm not sure about UB https://godbolt.org/z/KavPdbf3r

q4a: There is no unnamed structs in your case.
I can add names to structs and added it in some commits. Is it worth to add names in the `Color` class?

q4a: https://godbolt.org/z/8dcWfGY9j - here is more complex/real example. If remove `gcc_will_fail_if_removed` - gcc will fail.

q4a: @espkk Hi. Is there any news/problems to solve for this PR?

Hammie: I'm happy with these changes

espkk (approved):

PR #469, opened 2023-03-05
Only space to dialog name string if character has last name

### LINKED ISSUE #468: If there is no last name, then do not add a space in the dialog box. Ready code
Tiny kernel and script adjustments. As a result, we get this: if the NPC does not have a lastname (for example, it is an Indian), then there will be no space after the name in the dialog box.

1) In file "libs\dialog\src\dialog.cpp" 

You only need to edit one line. Below is the correct version of the function. I moved "m_sTalkPersName += " ";" one line below, in the inner part of the condition.

```
uint64_t DIALOG::ProcessMessage(MESSAGE &message)
{
    switch (message.Long())
    {
        // get character ID
    case 0:
        persId = message.EntityID();
        persMdl = message.EntityID();
        break;
        // get person ID
    case 1:
        charId = message.EntityID();
        charMdl = message.EntityID();
        const char *attr = nullptr;
        if (attr = core.Entity_GetAttribute(charId, "name"))
            m_sTalkPersName = attr;
        if (m_sTalkPersName.size() > 0)
        if (attr = core.Entity_GetAttribute(charId, "lastname"))
            m_sTalkPersName += " ";
            m_sTalkPersName += attr;
        m_BackParams.fCharacterNameRectCenterWidth =
            4.f +
            RenderService->StringWidth((char *)m_sTalkPersName.c_str(), m_nCharNameTextFont, m_fCharNameTextScale);
        break;
    }
    return 0;
}

```
2) In "program\dialog.c" add new if
	
```
	dialogRun = true;
	dialogSelf = false;
// start new if
	if(CheckAttribute(Character, "lastname"))
	{	
if(Character.lastname == "" || Character.lastname == "none")
	{
	DeleteAttribute(Character,"lastname");
	}
	}
// end new if
	LAi_Character_StartDialog(mainChr, Character);
	LAi_Character_StartDialog(Character, mainChr);
```

Hammie:
Fixes #468 

Adds a check so we only add the space in the dialog title when the characters last name is present and non-empty.

### DIFF
--- a/src/libs/dialog/src/dialog.cpp
+++ b/src/libs/dialog/src/dialog.cpp
@@ -1134,14 +1134,19 @@ uint64_t DIALOG::ProcessMessage(MESSAGE &message)
         charMdl = message.EntityID();
         const char *attr = nullptr;
         if (attr = core.Entity_GetAttribute(charId, "name"))
+        {
             m_sTalkPersName = attr;
-        if (m_sTalkPersName.size() > 0)
-            m_sTalkPersName += " ";
+        }
         if (attr = core.Entity_GetAttribute(charId, "lastname"))
+        {
+            if (m_sTalkPersName.size() > 0 && *attr != '\0')
+            {
+                m_sTalkPersName += " ";
+            }
             m_sTalkPersName += attr;
+        }
         m_BackParams.fCharacterNameRectCenterWidth =
-            4.f +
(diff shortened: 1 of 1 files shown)


Hammie:
@espkk Can you have a quick look? Should be fairly straight-froward.

espkk (approved):

PR #473, opened 2023-03-26
Nine changes

q4a:
It's draft for Gallium Nine support.


espkk:
We probably want to have separate conanfiles for linux and windows 🤔maybe after migration to conan 2.0

Hammie (dismissed):

q4a:
I would like to fix bug about trees without fog in Mesa: https://gitlab.freedesktop.org/mesa/mesa/-/issues/8341
Then I'll rebase on top of develop (move Nine support to `cmake/linux.cmake`).
And then it will be ready for merge.

q4a:
I rebased changed on top of develop and moved Nine support to `cmake/linux.cmake`. Looks like Mesa bug will not fixed soon. So I suggest to merge this changes now.

espkk (approved):

PR #478, opened 2023-04-14
Fix invalid lexicographical_compare

Hammie:
Fixes a theoretical issue because of using `lexicographical_compare` with `<=`. This triggered a debug assertion in Visual Studio.

Replaced it with the C++20 three-way comparison, which is hopefully supported on all compilers.

### DIFF
--- a/src/libs/util/include/istring.hpp
+++ b/src/libs/util/include/istring.hpp
@@ -1,0 +1,62 @@
+#pragma once
+
+#include <string>
+
+namespace storm
+{
+
+template <typename T> struct ichar_traits : public std::char_traits<T>
+{
+    static bool eq(T c1, T c2)
+    {
+        return std::toupper(c1) == std::toupper(c2);
+    }
+
+    static bool lt(T c1, T c2)
+    {
+        return std::toupper(c1) < std::toupper(c2);
+    }
+
+    static int compare(const T *s1, const T *s2, size_t n)
+    {
+        while (n-- != 0)
+        {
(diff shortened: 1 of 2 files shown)


q4a:
@espkk CI got `ERROR: Failed requirement 'directx/9.0@storm/prebuilt' from 'conanfile.py'`
Can you check it?

espkk:
Does clean local build work?

espkk:
Btw, don't we want to try implementing our char_traits instead?

Hammie:
`std::char_traits` could also work, I'll try it

Hammie:
I changed the code to use `char_traits`.
@espkk Is this sortof what you had in mind?

espkk:
> I changed the code to use `char_traits`. @espkk Is this sortof what you had in mind?

This looks good. Do we still have tests for it to make sure they are also fine? 

Now we can also do
```C++
using istring = std::basic_string<char, ichar_traits>;
```
and most importantly
```C++
using istring_view = std::basic_string_view<char, ichar_traits>;
```
and use it in most cases where we use those comparators. 

I'm not sure if we want to update all the existing code since it's pretty stable, but probably we can gradually simplify it in this way

espkk (approved):

Hammie:
We still have tests for `iEquals` and `iLess`. Didn't have to change those.

The aliases for `istring` and `istring_view` are also already in there.

Over time I think it would make sense to use those for our case-insensitive code and components. This will require us to do explicit conversions in between case-sensitive and case-insensitive code, but a more explicit border is probably a good thing.

PR #482, opened 2023-04-18
Add edit mode to LegacyDialog

Hammie:
- Refactor dialog link code, making the code more reusable
- Reuse this code in LegacyDialog, replacing its link handling code

The submitted edit value will now also be written directly to `Dialog.value`. For backwards-compatibility, value will still be written to `dialogEditStrings` as long as the `edit` attribute is not set to `"string"`.
This leaves the future possibilty of adding additional input modes or doing extra validation by specifying different edit types, ex.: `string`, `number`, `date`, etc...


q4a:
Error message: `dialog-test: error while loading shared libraries: libfmodL.so.13: cannot open shared object file: No such file or directory`
Problem is in Linux version of conan package `fmod`.

Hammie:
I don't think the tested code uses anything dialog-specific, so I'll probably just move the code to the utils library instead.

That way I won't have to deal with this dependency problem.

espkk (approved):

PR #483, opened 2023-04-18
[bsp-patcher] Fix sorting precision

espkk:


espkk:
todo: rewrite BSP completely

PR #485, opened 2023-04-21
Add `auto_pause` and `auto_mute` engine options to allow running in background

Hammie:
Adds two options to the `engine.ini`:

- `auto_pause`: Wether the game should be paused when the window loses focus. Default = `true`.
- `auto_mute`: Wether the audio should be muted when the window loses focus. Default = `false`, only checked when `auto_pause` is disabled.

### DIFF
--- a/src/apps/engine/src/main.cpp
+++ b/src/apps/engine/src/main.cpp
@@ -22,6 +22,7 @@ CorePrivate *core_private;
 constexpr char defaultLoggerName[] = "system";
 bool isRunning = false;
 bool bActive = true;
+bool bAutoMute = true;
 
 storm::diag::LifecycleDiagnosticsService lifecycleDiagnostics;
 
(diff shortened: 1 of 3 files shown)


src/apps/engine/src/main.cpp
```diff
@@ -95,7 +96,8 @@ void HandleWindowEvent(const storm::OSWindow::Event &event)
         if (core_private->initialized())
         {
             core_private->AppState(bActive);
-            if (const auto soundService = static_cast<VSoundService *>(core.GetService("SoundService")))
+            if (const auto soundService = static_cast<VSoundService *>(core.GetService("SoundService"));
```
espkk, line 99: oof.. we need to move this into sound service... but not as part of this PR of course

espkk:
how about something like `run_in_background` or `pause_in_background`, `sound_in_background`? it's not clear with `auto` that it relates to background

src/apps/engine/src/main.cpp
Hammie: Yes, very much agree

Hammie:
> how about something like `run_in_background` or `pause_in_background`, `sound_in_background`? it's not clear with `auto` that it relates to background

I think `run_in_background` and `sound_in_background` could indeed convey their purpose better. Can't hurt to have these options with a little more verbose names.

espkk (approved):
lgtm, thanks

PR #490, opened 2023-04-26
Fix nullptr exception during Fader init

### LINKED ISSUE #91: When talking to the Helmsman to go to sea, Game Crashes
In the beginning of the Horatio Hornblower questline, you need to speak to the Helmsman who you give directions to, and once you get through his dialogue, it's supposed to take you to the Sea View on the HMS Justinian, but instead the game crashes. Questline cannot continue.

Build: Alpha 8
Storyline: Horatio Hornblower

Crash logs and Save zipped together below. When loading the save, talk to the helmsman in front of you and exhaust his dialogue fully then wait a second, game will then crash.

[-=Player=- Antigua. Ship deck June 18th, 1798.zip](https://github.com/PiratesAhoy/new-horizons/files/11327649/-.Player.-.Antigua.Ship.deck.June.18th.1798.zip)

Hammie:
Fixes a nullptr exception.

Also took the opportunity to add a small template function for more easily getting Entities.

Should fix https://github.com/PiratesAhoy/new-horizons/issues/91

### DIFF
--- a/src/libs/core/src/entity_manager.cpp
+++ b/src/libs/core/src/entity_manager.cpp
@@ -203,6 +203,8 @@ void EntityManager::EraseAll()
 
     // clear cache
     cache_.Clear();
+
+    previousStamp_ = 0;
 }
 
 entptr_t EntityManager::GetEntityPointer(const entid_t id) const
@@ -315,12 +317,7 @@ entid_t EntityManager::CalculateEntityId(const size_t idx)
     // assert for idx exceeding 2^32
     assert(static_cast<size_t>(static_cast<uint32_t>(idx)) == idx);
 
-    // calculate stamp
-    const auto ms =
-        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch());
-
-    // assemble entity id
-    const auto stamp = static_cast<entid_stamp_t>(ms.count());
+    const auto stamp = ++previousStamp_;
     const auto id = static_cast<entid_t>(stamp) << 32 | idx;
 
     return id;
--- a/src/libs/core/src/entity_manager.h
+++ b/src/libs/core/src/entity_manager.h
@@ -71,7 +71,7 @@ class EntityManager final
     };
 
     static bool EntIdxValid(size_t idx);
-    static entid_t CalculateEntityId(size_t idx);
+    entid_t CalculateEntityId(size_t idx);
 
     size_t GetEntityDataIdx(entid_t id) const;
     entid_t GetEntityId(uint32_t hash) const;
@@ -95,4 +95,6 @@ class EntityManager final
 
     // stack of entities for deferred delete
     plf::stack<entity_index_t> deletedIndices_;
+
+    entid_stamp_t previousStamp_ = 0;
 };


espkk (commented):
You definitely shouldn't need a dynamic_cast there unless there is a logic error somewhere.
What is the issue we are trying to solve?

espkk:
Entity iterator should never return an invalid entity. If we have a crash there, it's either a bug in entity manager (unlikely) or that memory region is corrupted. Do you have a core dump for that issue?

Hammie:
The dynamic cast was just because clang keeps telling I should use it, and I think that is more correct. Not related to the actual issue.

I have two crash reports - that's how I figured it out myself - using the latest new-horizons engine (https://github.com/PiratesAhoy/new-horizons/releases/tag/15.0.0-alpha.8)

[Fader-nullptr.zip](https://github.com/storm-devs/storm-engine/files/11332315/Fader-nullptr.zip)

espkk:
> The dynamic cast was just because clang keeps telling I should use it, and I think that is more correct. Not related to the actual issue.
> 
> I have two crash reports - that's how I figured it out myself - using the latest new-horizons engine (https://github.com/PiratesAhoy/new-horizons/releases/tag/15.0.0-alpha.8)
> 
> [Fader-nullptr.zip](https://github.com/storm-devs/storm-engine/files/11332315/Fader-nullptr.zip)
> 

Clang doesn't know that we have a strong type guarantee, so we should never use dynamic_cast with entities because it's essentially a pointless overhead 

I will check the dump later on today

PS I remember about the sentry issue, there is something wrong on their side, I will try to add you manually

espkk:
@Hammie, I cannot find matching symbols for those dumps, can you send me a link?

Hammie:
I think these are the ones used?
[alpha.8-engine.zip](https://github.com/storm-devs/storm-engine/files/11334161/alpha.8-engine.zip)

espkk:
> I think these are the ones used? [alpha.8-engine.zip](https://github.com/storm-devs/storm-engine/files/11334161/alpha.8-engine.zip)

Thanks, I didn't realize the binaries are included in the release you mentioned. I can see the issue, but as I said before, in theory this should never happen because `GetEntityIds` returns only valid entities... Do you think we can reproduce this?

Since this looks like a bug I would rather try to fix it than adding a workaround, otherwise it may appear in other places

Hammie:
I was not able to reproduce it from the save games, so it likely is undefined behavior at runtime.

I could make a build with a lot of extra checks for the entity manager, so we can maybe find the source of the issue?

espkk:
> I was not able to reproduce it from the save games, so it likely is undefined behavior at runtime.
> 
> I could make a build with a lot of extra checks for the entity manager, so we can maybe find the source of the issue?

Yeah, that could help. We probably need a printout functionality for entity manager to trigger it manually when we detect invariant violation, or even to trace entities' lifecycles.
I had something like that when I was debugging similar issues, but unfortunately I cannot find that branch...it's probably on my work laptop

Hammie:
I added some crude `EntityContainerCache` validation to the code, and I think I have found the problem.

The cache assumes that it stores all entity ids in a sorted order (low to high), but this is not always the case:
afbeelding

Apparently entity ids use a timestamp in the higher 32-bits. This timestamp is derived from the number of milliseconds since epoch: 
https://github.com/storm-devs/storm-engine/blob/fee388293aadc6fc690a28371ca0b787faea7c36/src/libs/core/src/entity_manager.cpp#L319-L320

Problem 1 is that **milliseconds** since epoch does not fit in 32-bits, and loops (every ~50 hours?)
Problem 2 is that multiple entities can be created during the same millisecond

The cache assumes all created entities to have ascending ids:
https://github.com/storm-devs/storm-engine/blob/fee388293aadc6fc690a28371ca0b787faea7c36/src/libs/core/src/entity_container_cache.cpp#L19-L20

Which means that some entities will not be correctly removed from the cache.

Hammie:
Replacing the call to `EntityContainerCache::UpdateAdd` with `EntityContainerCache::Add` should fix the issue.

But then I wonder why `EntityContainerCache::UpdateAdd` was ever added?

espkk:
Oof... this is really surprising, but the root cause is not clear yet

> Problem 1 is that milliseconds since epoch does not fit in 32-bits, and loops (every ~50 hours?)
> Problem 2 is that multiple entities can be created during the same millisecond
>
Those are not really problems. Originally, there were more data to distinguish entities but entity_id had much bigger size and a few additional levels of indirection. When I was reworking entity management I was trying to make it as cache friendly as possible.
So, my idea was to create a unique entity id that is very unlikely to collide. 
Using least 32bits of timestamp + 32 bit incrementing cookie means that once the timestamp is "overflowed" there should be approx. 1/2^32 chance to have a collision with another entity. This is not very accurate but should be reasonable enough to be confident about possible collisions. At least, I still don't see any inconsistency there.

The cache was a later addition to make it even more cache friendly and decrease "named" lookup time significantly. Replacing `UpdateAdd` with `Add` would make it logN times slower. But once again, it should not be a problem if the logic is correct.

In other words, we always assume that higher 32 bits of entity id are not decreasing, while least 32 bits are strongly increasing, that's why adding new entity should never result in an unsorted sequence like you got on the screenshot.

So, it sounds like a bug and we should try to find it

espkk:
Wait.. did we use least 32bits for indexing? And then probably there is a free-indices stack? Then it makes sense...it's not correct
I'm not sure what would be better, removing named-caching completely or preventing this with e.g. banning indices stack usage within the same millisecond.
I would prefer the latter, but the former should be easier to implement

Hammie:
Even if we get rid of the free indices stack, there is still the issue that the stamp will loop, and thus is not guaranteed to always increase. 
Also discovered that if you start the game just when the clock resets, and the stamp == 0, the first entity you create will be considered invalid (stamp == 0, index ==0).
So I think we should get rid of the timestamp altogether.

Hammie:
I've replaced the timestamps with a simple counter, and that fixes the issue. No more duplicates, and is guaranteed to always increase.

Also found that we need to delete the entity before marking it deleted in `EntityManager::EraseAll`, as entities can delete other entities.

espkk:
You mean replace timestamp with simply incrementing cookie? But isn't it just the same but time amount needed to overflow is just a bit larger? Also, timestamp is very useful when debugging lifetimes, but we can keep it inside internal entity data under the flag (not in the entity itself, but in its node inside entity manager)
In both cases 0/0 can be avoided by simply overjumping zero. But I'm not sure I got the idea

espkk:
> Also found that we need to delete the entity before marking it deleted in EntityManager::EraseAll, as entities can delete other entities.

I don't think I understand that one either

espkk:
Regarding invalid_entity we should probably have it -1 instead, that would mean entity 2^32 index which is in fact not possible

Hammie:
The problem with the timestamp is that it uses the system clock, not time since engine start. So every 50 days (`4,294,967,296 / (3,600,000 * 24) = 49.7103`) everyone's game will start crashing at the same time.

If we use a counter we should be good until someone creates 2^32 entities without restarting or reloading.

Hammie:
> > Also found that we need to delete the entity before marking it deleted in EntityManager::EraseAll, as entities can delete other entities.
> 
> I don't think I understand that one either

When marking the entity as `kNotExists` before deleting it. It will still be referenced in the caches while deleting other entities, causing the caches to contain "invalid" entities.

espkk:
> The problem with the timestamp is that it uses the system clock, not time since engine start. So every 50 days (`4,294,967,296 / (3,600,000 * 24) = 49.7103`) everyone's game will start crashing at the same time.
> 
> If we use a counter we should be good until someone creates 2^32 entities without restarting or reloading.

yeah... the epoch timestamp is not good unless we are giving up with "only increasing" assumption for our "cache". But do we have any measurements of entities amount e.g. while f6/f9? if it's just thousands or tens of thousands then we should be OK with just incrementing cookie

>When marking the entity as kNotExists before deleting it. It will still be referenced in the caches while deleting other entities, causing the caches to contain "invalid" entities.
>
Then we can just clear the cache before starting destroying entities. entity dtors shouldn't have any side effects via other entities so since we are destroying everything we should be safe. We could also add nullptr checks while traversing returned iterator. But do we even have `GetEntityByName` in dtors?

UPD
Let me take a closer look in the next few days. That would probably be more effective than my quick but not very helpful answers.
And also thanks for investigating this, I was sure we don't have any issues there anymore

Hammie:
I loaded a savegame in New Horizons, and the stamp was only at ~150 entities created, so I think it should last a very long time.

Added a local debug assertion when it goes over 10.000, to see how quickly I'll hit it.

espkk:
I think we're fine merging this except for the exceptions - entity manager should not throw
I would rewrite most of its code completely but it can take some time while we want to fix the issues ASAP

Hammie:
I've removed everything except the actual fix

espkk (approved):

PR #491, opened 2023-05-03
Allow free camera to be used when not near an island

Hammie:
Allow free camera to be used when not near an island.

Also allow free camera to move when numlock is enabled. Not sure why this check existed?


espkk (approved):
LGTM

IIRC, it centered on the island when at sea, probably there was a limitation during development
The same was for num_lock, camera switch was not implemented yet

PR #492, opened 2023-05-05
Fixed perspective on ultrawide screens

Reksotiv:
In the **SetPerspective** function, the variable **perspective** is the horizontal fov, which is set to the same value for all diagonals, and as the screen width increases, the image gets closer. For example, the horizontal fov is 90 and with an aspect ratio of 16:9 vertical fov = 59, and with an aspect ratio of 21:9 vertical fov = 47. https://themetalmuncher.github.io/fov-calc/
The perspective can be increased using the **fov_multiplier** variable in the **engine.ini** file, but after that the image looks distorted. Look at the screenshot with the barrel, without changes it is not round but oval.
Instead of manually calculating perspective projection matrix, i used this function https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dxmatrixperspectivefovlh

**Without fix fov_multiplier = 1.0**
default-1 0
**With fix fov_multiplier = 1.0**
moded-1 0

**Without fix fov_multiplier = 1.5**
default2-1 5
**With fix fov_multiplier = 1.5**
moded2-1 5

**Without fix fov_multiplier = 1.5**
default-1 5-circle
**With fix fov_multiplier = 1.5**
moded-1 5-circle

### DIFF
--- a/src/libs/renderer/src/s_device.cpp
+++ b/src/libs/renderer/src/s_device.cpp
@@ -2094,30 +2094,17 @@ void DX9RENDER::SetNearFarPlane(float fNear, float fFar)
 
 bool DX9RENDER::SetPerspective(float perspective, float fAspectRatio)
 {
-    perspective *= FovMultiplier;
+    if (fAspectRatio < 0)
+    {
+         fAspectRatio = static_cast<float>(screen_size.x) / screen_size.y;
+    }
 
     const float near_plane = fNearClipPlane; // Distance to near clipping
     const float far_plane = fFarClipPlane;   // Distance to far clipping
-    const float fov_horiz = perspective;     // Horizontal field of view  angle, in radians
-    if (fAspectRatio < 0)
-    {
-        fAspectRatio = static_cast<float>(screen_size.y) / screen_size.x;
-    }
-    aspectRatio = fAspectRatio;
-    const float fov_vert =
-        2.f * atanf(tanf(perspective / 2.f) * fAspectRatio); // Vertical field of view  angle, in radians
+    const float fov_vert = 2.0f * atanf(tanf(perspective * 0.5f * FovMultiplier) / fAspectRatio);
 
-    const float w = 1.0f / tanf(fov_horiz * 0.5f);
-    const float h = 1.0f / tanf(fov_vert * 0.5f);
-    const float Q = far_plane / (far_plane - near_plane);
-
-    D3DMATRIX mtx{};
-
-    mtx._11 = w;
-    mtx._22 = h;
-    mtx._33 = Q;
-    mtx._43 = -Q * near_plane;
-    mtx._34 = 1.0f;
+    D3DXMATRIX mtx{};
+    D3DXMatrixPerspectiveFovLH(&mtx, fov_vert, fAspectRatio, near_plane, far_plane);
 
     if (CHECKD3DERR(d3d9->SetTransform(D3DTS_PROJECTION, &mtx)) == true)
         return false;
--- a/src/libs/renderer/src/s_device.cpp
+++ b/src/libs/renderer/src/s_device.cpp
@@ -2094,17 +2094,28 @@ void DX9RENDER::SetNearFarPlane(float fNear, float fFar)
 
 bool DX9RENDER::SetPerspective(float perspective, float fAspectRatio)
 {
-    if (fAspectRatio < 0)
-    {
-         fAspectRatio = static_cast<float>(screen_size.x) / screen_size.y;
-    }
-
     const float near_plane = fNearClipPlane; // Distance to near clipping
     const float far_plane = fFarClipPlane;   // Distance to far clipping
-    const float fov_vert = 2.0f * atanf(tanf(perspective * 0.5f * FovMultiplier) / fAspectRatio);
+    const float fov_horiz = perspective;     // Horizontal field of view  angle, in radians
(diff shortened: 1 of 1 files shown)


Hammie:
I think I understand the issue, and `const float fov_vert = 2.0f * atanf(tanf(perspective * 0.5f * FovMultiplier) / fAspectRatio);` is the actual part of the fix we need.

Let me double check the math.

Hammie (changes requested):
The vertical fov calculation is indeed incorrect. This could should fix it:

``` 
perspective *= FovMultiplier;

const float near_plane = fNearClipPlane; // Distance to near clipping
const float far_plane = fFarClipPlane;   // Distance to far clipping
const float fov_horiz = perspective;     // Horizontal field of view  angle, in radians
if (fAspectRatio < 0)
{
    fAspectRatio = static_cast<float>(screen_size.y) / screen_size.x;
}
aspectRatio = fAspectRatio;
const float fov_vert = 2.f * atanf(tanf(perspective / 2.f) * fAspectRatio); // Vertical field of view  angle, in radians

const float w = 1.0f / tanf(fov_horiz * 0.5f);
const float h = 1.0f / tanf(fov_vert * 0.5f);
const float Q = far_plane / (far_plane - near_plane);

D3DMATRIX mtx{};

mtx._11 = w;
mtx._22 = h;
mtx._33 = Q;
mtx._43 = -Q * near_plane;
mtx._34 = 1.0f;
```

Only the `fov_vert` calculation needs updating

Reksotiv:
You are right, only fov_vert

Hammie (dismissed):

Reksotiv:
No, changing only fov_vert is not enough, sea camera not working properly. I made this change 3 months ago and forgot why i changed D3DMATRIX math.
**Only fov_vert**

Hammie:
What game/mod are you playing? I'm not seeing any issues when I set TEHO to a 21:9  aspect ratio.

Reksotiv:
increase **fov_multiplier** in **engine.ini**, i use 1.35 for 3440x1440 resolution. Any game, screenshots above from beyond new horizons, but all tests were in a ship pack 2.3.0a. With the change played 50 hours without problems and 10 hours in TEHO.
Without using the **D3DXMatrixPerspectiveFovLH** function, i remember changing something in the files deck_camera.cpp, free_camera.cpp, ship_camera.cpp. But why don't you want to use it? I don't know something or?

Hammie:
> But why don't you want to use it? I don't know something or?

We are trying to depend less on DirectX, for cross-platform compatibility. Adding new dependencies is something I would really like to avoid.

Reksotiv:
I understand why the problem with the camera in the sea. 
image
in my fix i removed that and applied the FovMultiplier without changing the perspective variable, which is stored in [Fov variable here.](https://github.com/storm-devs/storm-engine/blob/develop/src/libs/renderer/src/s_device.cpp#L2123)
image
and when we move from land to sea, we get a double FovMultiplier

espkk:
> > But why don't you want to use it? I don't know something or?
> 
> We are trying to depend less on DirectX, for cross-platform compatibility. Adding new dependencies is something I would really like to avoid.

Just to clarify, that's not about DirectX itself, but rather about D3DX support lib which is not present for our Linux build, so it simply won't compile unless we implement it ourselves (should be a few lines though)

> I understand why the problem with the camera in the sea. image in my fix i removed that and applied the FovMultiplier without changing the perspective variable image and when we move from land to sea, we get a double FovMultiplier [ship_camera.cpp](https://github.com/Reksotiv/storm-engine/blob/develop/src/libs/sea_cameras/src/ship_camera.cpp#L41)

Yes indeed, IIRC it was my attempt to centralize FOV customization, but it ended up setting it through sea/camera only for compatibility. Also, it's probably still missing from the SKY entity. So `fov_multiplier` was an unfinished feature.

With `fov_multiplier=1.0` it likely won't affect anything even if we merge it as is, but since we are touching this we probably want it to be complete. And this is a non-trivial task...

Reksotiv:
this code works fine for me, without saving the changed perspective by fov_multiplier in Fov variable

    const float near_plane = fNearClipPlane; // Distance to near clipping
    const float far_plane = fFarClipPlane;   // Distance to far clipping
    const float fov_horiz = perspective;     // Horizontal field of view  angle, in radians
    if (fAspectRatio < 0)
    {
        fAspectRatio = static_cast<float>(screen_size.y) / screen_size.x;
    }
    aspectRatio = fAspectRatio;
    // Vertical field of view  angle, in radians
    const float fov_vert = 2.f * atanf(tanf(perspective * 0.5f * FovMultiplier) * fAspectRatio);

    const float w = 1.0f / tanf(fov_vert * 0.5f) * fAspectRatio;
    const float h = 1.0f / tanf(fov_vert * 0.5f);
    const float Q = far_plane / (far_plane - near_plane);

    D3DMATRIX mtx{};

    mtx._11 = w;
    mtx._22 = h;
    mtx._33 = Q;
    mtx._43 = -Q * near_plane;
    mtx._34 = 1.0f;

Reksotiv:
> Yes indeed, IIRC it was my attempt to centralize FOV customization, but it ended up setting it through sea/camera only for compatibility. Also, it's probably still missing from the SKY entity. So `fov_multiplier` was an unfinished feature.

ok i realized that with this fix we might get errors in other places.

PR #508, opened 2023-09-24
Fix IsEntity sometimes returning true for previously deleted entities

### LINKED ISSUE #141: Controls not working when loading sea view
When saving the game in sea view, then loading it, none of the controls work (except F1 - main menu).

Hammie:
When deleting an entity, saving the game and then loading that save game, it is possible for a newly created entity to have the same id as the old deleted entity id.

If the old entity variable is not yet overwritten, it will then return true for `IsEntity(&entity)`.

This can be fixed by making sure the variable value is reset after calling `DeleteEntity`

Also fixes https://github.com/PiratesAhoy/new-horizons/issues/141

### DIFF
--- a/src/libs/core/src/internal_functions.cpp
+++ b/src/libs/core/src/internal_functions.cpp
@@ -1626,6 +1626,10 @@ DATA *COMPILER::BC_CallIntFunction(uint32_t func_code, DATA *&pVResult, uint32_t
         }
         pV->Get(ent);
         core.EraseEntity(ent);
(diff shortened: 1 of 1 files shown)


ugeen4 (approved):

espkk:
While it helps in some cases it's not the root cause and has to be fixed differently...
So, we probably don't want to merge this. Instead, we shouldn't zero the cookie counter during entity manager's full-erase

PR #511, opened 2023-11-21
Fix dynamic call screwing up values

mitrokosta:
ExpressionResult may contain a reference to an object after returning from BC_CallFunction. Script example that insta crashes the game before this fix (most of the time, sometimes it hides MC's portrait):
```
void Suicide()
{
	ref sld = CharacterFromID("Blaze");
}

void ActiveF10Control()
{
	string death = "Suicide";
	call death();
}
```
`Suicide()` caused ExpressionResult to have a reference to an object (reference to main character). `ExpressionResult.Set(0);` sets the object's type to int which is obviously bad.


q4a:
Linux fail because of Clang 12 removed from Ubuntu 22.04 CI image: https://github.com/actions/runner-images/issues/8263
I'll try to fix CI today.

